PR #23, opened 2017-04-04
Remove facility caching

ratneshraval:
Removing facility caching till we have a caching strategy, leaving helper class as is


ratneshraval:
This doesn't work for AWS deployment. Rather than updating AWS to support this POC, I removed it so we can continue with other stories and take this one on when we have plan for caching and deploying it.

lauherkcambria:
ok, that makes sense. we will disable the caching for now until we have a clearer caching strategy.  We will leave the fetch_facilities method, however, as the method will most likely be used once we have a strategy in place.

PR #24, opened 2017-04-04
Redis host in ENV vars. Also switching to yarn. 

ratneshraval:


ratneshraval:
This switches to yarn when installing in AWS.
Also adding Redis host in docker file.

PR #26, opened 2017-04-05
Dockerfile updated to use yarn for packages

ratneshraval:
Missed the line where we are using npm install in docker file.
Updated it to use yarn


lavanyareddyreddy:
👍

PR #30, opened 2017-04-11
changing assets port to resolve jenkins conflict

ratneshraval:
### Brief feature description:
since jenkins uses port 8080 we need to use 5080 for webpack 

### Does this PR block/effect other stories?
N/A - no stories are affected by this implementation.

### Which User story does this relate to?
Story-> #140420789

### Do any commands needs to be run in order to run the project with this PR?
N/A

### Code coverage percent diff as a result of this PR?
N/A

### JS errors diff as a result of this PR?
N/A


ratneshraval:
Marco!

lauherkcambria:
@ratneshraval polo

PR #31, opened 2017-04-13
db2 tables creation scripts

ratneshraval:
* #142793245
* table creation scripts for cws/cms tables
* No
* No
* No
* No


lavanyareddyreddy:
👍

PR #33, opened 2017-04-17
fixing composite PK column names

ratneshraval:
- #142793245

- Postgres script didn't like all caps for composite key names. Lowering the case

- No

- No 

- N/A

- N/A


lavanyareddyreddy:
👍

PR #34, opened 2017-04-24
Karma coverage configuration update with js files and spec files

anuroop15:
*Brief feature description

    Karma coverage configuration update with js files and spec files
*Does this PR block/effect other stories?

    n/a
*Which User story does this relate to?

    143917787
*Do any commands needs to be run in order to run the project with this PR?

    yarn add istanbul-instrumenter-loader
*Code coverage percent diff as a result of this PR?

    n/a
*JS errors diff as a result of this PR?

    n/a


ratneshraval:
@anuroop15  Please update PR request to match the format. 

Command needed to run this 
    yarn add istanbul-instrumenter-loader

PR #38, opened 2017-05-01
update facility search to use Facility concern

ratneshraval:
- N/A

- Updating facility model to use search from facility concern module

- No

- No. 

- n/a

- n/a


ratneshraval:
Feelings < Action  *boom*

ratneshraval:
@lauherkcambria we can take it up as separate story. This PR is for bug fix and will not refactor

PR #39, opened 2017-05-01
Env vars

ratneshraval:
- #143902253

- Extracting out environment variables and moving to a single read file

- No

- No. But keep your .env file up to date with all variables for development. 

- n/a

- n/a


ratneshraval:
All variables needed should be in that file and wiki. Currently only 4.

PR #40, opened 2017-05-01
Sp2.4 remove search criteria

anuroop15:
-#143205713

-Facility search criteria can be changed by removing the selected criteria 

-No

-yarn install  (A new package was added "prop-types")
-npm run dev

- increased from 75% to 77%

- decreased error count from 600 579.


anuroop15:
@ratneshraval - Updated with added Package

PR #41, opened 2017-05-01
add v1 of complaints table. set up similar to child table

lauherkcambria:
…data to be replaced with API

- https://www.pivotaltracker.com/story/show/143985031
-Adding a complaints table to the facility profile page
-N/A
-N/A
- 67% 
- 647


lauherkcambria:
brought down the lint errors to 33

.eslintrc
```diff
@@ -1,5 +1,5 @@
 {
-"extends": "airbnb",
+"extends": "standard",
```
anuroop15: Any Reason why we are going to standard instead of airbnb as discussed in our dev meetings.

anuroop15:
@lauherkcambria : Any Reason why we are going to standard instead of airbnb as discussed in our dev meetings.

PR #43, opened 2017-05-02
Back button landing page

anuroop15:
-#143986289

- CALS can navigate back to landing page by clicking back button

- No

- npm run dev

- No Changes: 75% -> 75%

- No changes: 625 -> 625


anuroop15:
Added as as test variable. will remove that.

PR #49, opened 2017-05-08
sending cookies with fetch

ratneshraval:
- #144644377

- sending cookies with fetch, so it honors authenticated session

- No

- No

-  N/A

- N/A


lavanyareddyreddy:
The "same-origin" value makes fetch behave similarly to XMLHttpRequest with regards to cookies. 

https://github.com/github/fetch#sending-cookies

Otherwise, cookies won't get sent, resulting in these requests not preserving the authentication session.

PR #52, opened 2017-05-09
heartbeat page

ratneshraval:
- #144495343

- heartbeat check page that bypasses authentication check

- No

- No

- 42%

- 45 -> 45


ratneshraval:
Good point. I'll update the test.

lauherkcambria:
from chat with nesh: 

In a controller spec you check the response, in a request spec, you check the page content, since you have only access to the html.

PR #53, opened 2017-05-10
elasticsearch query builder

ratneshraval:
- Yes

- https://www.pivotaltracker.com/story/show/143904853

- Building elastic search query from search hash

- This story was marked as dependency on search facility page update. This will help to retrieve accurate search results

- no

- Spec 74%

- 21

### DIFF
--- a/spec/classes/elastic/query_builder_spec.rb
+++ b/spec/classes/elastic/query_builder_spec.rb
@@ -1,0 +1,54 @@
+require 'rspec'
+require 'rails_helper'
+
+describe FacilitiesController do
+
+  describe "Build query" do
+    it "builds simple bool AND" do
+
+      expected_output = {
+        bool: {
+          must: [
+            { match: {'fac_co_nbr':'28'} },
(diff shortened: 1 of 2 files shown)


spec/classes/elastic/query_builder_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+                  {match:{'fac_co_nbr':'18'}}
+                ]
+                }
+               }]
+          }
+        }
+      }
+
+      input = [ {fac_co_nbr: '28', fac_name: 'home'}, {fac_co_nbr: '18'} ]
+      output = Elastic::QueryBuilder.match_boolean(input)
```
lavanyareddyreddy: as per ES query , does the input here supposed to be [ {fac_co_nbr: '28', fac_name: 'home'}, {fac_co_nbr: '18', fac_name: 'home'} ]?

ratneshraval: The input can be whatever you want it to be. This just demonstrates a sample use case.

PR #54, opened 2017-05-10
Sp2.5 link profile to search

anuroop15:
-Javascript tests for profile page links in search page. All tests passed

- #144050395

- Ability lo load Facility profile page from search results. 

-No

-npm run dev

-JS 90.1%, rspec: 72.21%

-21


ratneshraval:
Let's pair and review this tomorrow.

anuroop15:
@lauherkcambria 
1) Nice thought on rendering mock data as a factory. I initially thought of keeping it separate so that we can use it a singleton object. I will refactor it going further in other US's.
@2: Not user on how the UX design goes. but if there wasn't a back button, we will make search results saved when browser back button was clicked. Which comes in separate User story.

lauherkcambria:
@anuroop15 awesome, thanks for your reply! 👍 

re: @2 what do you mean by 'we will make search results saved when browser back button was clicked'? 

I just thought this will most certainly become a requested feature later on, so it may be a good idea to at least start thinking about it now.

anuroop15:
re:@2  I mean browser default back button. 
correct!! I'm checking the best method to handle back button.

PR #59, opened 2017-05-15
Facility search to use es query

lavanyareddyreddy:
-yes
92.42% Statements 61/66 0% Branches 0/2 86.36% Functions 19/22 92.42% Lines 61/66

-#143905057

- search to use new api endpoin call(post).
- search page should build elasticsearch query

- no

- npm run karma 

- n/a

-n/a


ratneshraval:
1. field names coming in are not matching with fields in the index (county vs fac_co_nbr).
2. Test coverage is too low for added code.

app/controllers/facilities_controller.rb
```diff
@@ -2,33 +2,32 @@
 
 class FacilitiesController < CalsBaseController
   around_action :hypernova_render_support
+  skip_before_action :verify_authenticity_token, only: [:search]
```
ratneshraval: Let's figure out how we can remove this line and keep "verify_authenticity_token"

lavanyareddyreddy:
1. Field names wont match because once elastic search returns result we use serializer.
2. updated with spec tests.

app/controllers/facilities_controller.rb
lavanyareddyreddy: we need to set CSRF token in JS and then we can remove skip_before_action.

ratneshraval:
@lavanyareddyreddy  can you explain 1? I meant request field names are not matching. Which returns inaccurate results.

lavanyareddyreddy:
@ratneshraval : updated county and type column names to match field names (fac_co_nbr, fac_type). please check my commit 35651b0

anuroop15:
why not ? We don't have separate US for this refactoring. So I committed those changes here. I'm writing some testcases for this US and thought of modifying the folder structure.

lauherkcambria:
if there isn't a user story for this, shouldn't we make one? I feel that it's not a good call putting in unrelated commits to a specific feature branch

ratneshraval:
@anuroop15 you don't need story to modify the structure. You can add chore if you need to. I recommend that you only modify structure of files you need to add test cases for this branch/story and rest of them modify when you take up the chore in a separate branch.

lauherkcambria:
@ratneshraval wouldn't it be better to have all structure modifications @anuroop15 is doing be apart of one commit that is on a PR related to a single chore? more granularity that way

ratneshraval:
What Keith said works too. I didn't propose that option in case it becomes double work. You can write tests with current structure and then re modify structure of all in new PR.

anuroop15:
cool!! I will create a chore for refactoring and keep separate PR.

ratneshraval:
Ruby code coverage dropped below 90%. Something to consider and improve.

lauherkcambria:
@lavanyareddyreddy @ratneshraval nice!!!! before we merge this lets sit down tomorrow morning and get the coverage back to 100 👍

ratneshraval:
Please pull dev branch in here so you don't have to recreate tests Keith already added.

lavanyareddyreddy:
@ratneshraval @lauherkcambria - merged Development branch, fixed failing tests, code coverage shows 100%.

lauherkcambria:
noice 👍 🥝

PR #65, opened 2017-05-17
refactor assets for POC. add test cases, refactor out unused files

lauherkcambria:
- rspec search facility controller test was added, i noticed it wasn't there, all passing

-  #145622391

- this is a proof of concept showing how we can take assets out of our javascript folders into a shared space as necessary -- this will help optimize css bundling ultimately helping performance

- yes, refactoring of all assets, #145623409

- npm run dev-build; npm run all-tests;

- 92 -> 100

- 21 -> 2


anuroop15:
Nice work @lauherkcambria 
Just adding my thoughts:
1) I feel stylesheets in two different locations was inconvenient. If we can render from asserts folder by importing, can we have the "assets/stylesheets/facilities" &  "assets/stylesheets/search".. etc along with  "assets/stylesheets/shared"
2) It would be good to have shared folder for images too.

ratneshraval:
@anuroop15 
1. 2 difference places makes it easy to identify what is shared and what is custom/specific to the module. Even if we put it in the same location, we will still have to have a way to distinguish what is custom to the module. I suggest we go with this hybrid approach.
2. Images will be moved to common location and will be used in stylesheets like what I have in my last commit 721e43b

PR #67, opened 2017-05-19
fix dropdown issue to open list

anuroop15:
-NA

-#145471745

- Dropdown having down arrow made clickable and can open dropdown list

-No

-npm run dev

- rspec - 100%, JS - 90.24%

- error => 1


ratneshraval:
Looks good. As per our discussion we need to move the webkit contact autofill style into the shared section at some point. Maybe add a chore for that.

lauherkcambria:
👍

PR #68, opened 2017-05-23
fix ie11

lauherkcambria:
- N/A

- https://www.pivotaltracker.com/story/show/145470913

- fix search not working in IE 11

- no

- no

- N/A

- N/A

### DIFF
--- a/config/webpack/shared.js
+++ b/config/webpack/shared.js
@@ -37,7 +37,8 @@ module.exports = {
   plugins: [
     new webpack.EnvironmentPlugin(JSON.parse(JSON.stringify(env))),
     new ExtractTextPlugin(env.NODE_ENV === 'production' ? '[name]-[hash].css' : '[name].css'),
-    new ManifestPlugin({ fileName: paths.manifest, publicPath, writeToFileEmit: true })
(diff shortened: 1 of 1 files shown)


ratneshraval:
Would replacing whatwg-fetch with -polyfill version fix it? And help us for such future issues of fetch?
Or do we need to add this new package.

lauherkcambria:
this approach allows us to include the plugin in one place and work for all of our javascript modules. I think this approach is easiest and most efficient. My choice for including this plugin rather than a polyfill is that it would be less refactoring.

lauherkcambria:
that said though, if you would like to explore other ways of fixing this bug by all means go for it! I think your question is more along the lines of, 'should we be really using whatwg-fetch instead of something better that doesn't break IE11?'

lauherkcambria:
in which case, i believe that research should be it's own user story, and this bug should be approved in the meantime as a fix for our current setup.

ratneshraval:
I say we go with this fix. If we run into another issue with fetch then we can create a spike. Otherwise just live with this fix.

lauherkcambria:
sounds good to me 😆

PR #69, opened 2017-05-24
headless browser testing via phantom and poltergiest. BOO!

lauherkcambria:
scary, BOO~~!

- no tests were added as this  is to upgrade our test suite to use browserless testing instead of chrome so it can be used with CI / jenkins. 

-  jenkins stories

- switch from chrome to phantomjs for testing

-

- yarn && bundle update

- same 

- N/A


ratneshraval:
Current error on Jenkins server 
Could not find an executable 'phantomjs' that matched the requirements '>= 1.8.1', '< 3.0'. Found versions were {"/var/lib/jenkins/workspace/PR testing - CALS/node_modules/.bin/phantomjs"=>"294"}.

lauherkcambria:
now that we removed awesome print and byebug and got errors fixed on jenkins, @ratneshraval @anuroop15 @lavanyareddyreddy , does this PR look good to merge?

ratneshraval:
What's our plan to remove auth check in feature tests?

lauherkcambria:
@ratneshraval  we found a fix by replacing that auth check with the following:

`allow_any_instance_of(CalsBaseController).to` receive(:authenticate_with_cwds).and_return(true)`
 
this will bring code . coverage down to 95%, but we aren't 100% sure about how to get around this so we will make a user story to research and bring coverage back up to 100.

ratneshraval:
Ruby coverage dropped to 95%. We need to create a separate story and bring that up to 100%.
Otherwise it looks good.

PR #78, opened 2017-06-05
remove auth test and clear some lint errors

ratneshraval:
- Auth test removed since it cannot function in Jenkins. Needs rewrite as unit test instead of feature.

- https://www.pivotaltracker.com/story/show/145482647

- Auth test removed since it cannot function in Jenkins. Needs rewrite as unit test instead of feature.

- Yes. All deploys blocked.

- No

- 100 -> 96%

- 0


ratneshraval:
retest this please

PR #79, opened 2017-06-05
fix unresponsive table for children and complaints

lauherkcambria:
- n/a

-#146491781

- fix the children and complaints table for less than 600px

- no

- no

- 100

- 0


ratneshraval:
/public files should not be committed. Can you remove them?

app/javascript/facility/stylesheets/facility.scss
```diff
@@ -107,6 +107,46 @@
     color: #2b2b2b;
 }
 
+@media screen and (max-width: 600px){
```
anuroop15: @keith, are you targeting mobile screens? As we are following Bootstrap guidelines, mobile screens should be 767px and below. Is there aa . reason you target upto 600 only?

keith: Wrong Keith 😄

lauherkcambria: apologies @keith! 🙇 🙏

lauherkcambria: @anuroop15, i set it as 600 as I thought that was the standard -- can be corrected to 767, i was going off of the standard: 
`@media (min-width:600px) and (max-width:767px) {}`

which is probably why. i'll update and please approve. thanks

PR #86, opened 2017-06-14
React rails

ratneshraval:
- No tests added. 

- N/A

- Switching to react-rails instead of hypernova.

- No

- Bundle install. Yarn install.

- 100%

- 0


ratneshraval:
There is 1 lint error in a sample file generated (hello_react.jsx). We can remove that file or fix the error.

lauherkcambria:
good call @ratneshraval , i removed the unused jsx file and reformatted test.js to be in standard.js format 🌮

app/controllers/facilities_controller.rb
```diff
@@ -1,19 +1,19 @@
...
   def index
-    @facilities = facility_helper.all.to_json
+    @facilities = facility_helper.all
   end
 
   def show
     @facility = facility_helper.find_by_id(params[:id])
     # @facility ||= Facility.find_by_id(params[:id])
-    @children ||= @facility.children.to_json
-    @complaints ||= @facility.complaints.to_json
```
lauherkcambria: i like that we don't have to do to_json everywhere

config/initializers/hypernova.rb
```diff
@@ -1,4 +1,4 @@
-Hypernova.configure do |config|
```
lauherkcambria: i think we should remove this file

app/controllers/facilities_controller.rb
lauherkcambria: i like that we don't have to do to_json everywhere

spec/controllers/facilities_controller_spec.rb
```diff
@@ -11,7 +11,7 @@
   describe 'GET index' do
     it 'assigns @facilities' do
       get :index
-      expect(assigns(:facilities).size).to eq(38547)
+      expect(assigns(:facilities).size).to eq(53)
```
lauherkcambria: this is cleaner too as a result, counting  whole objects rather than bits of json

app/views/facilities/index.html.erb
```diff
@@ -1,5 +1,5 @@
 <div id = 'root'>
-    <%= render_react_component('facilities', facilities: @facilities) %>
+    <%= react_component('facilities', {facilities: @facilities}, {prerender: true}) %>
```
lauherkcambria: i'm concerned about prerender true everywhere -- i'm wondering if side-effects of the prerender will affect later renders with that renderer according to docs. we may in the future have to use  per_request_react_rails_prerenderer as described in https://github.com/reactjs/react-rails#javascript-state

app/javascript/packs/search.js
```diff
@@ -1,8 +1,7 @@
...
-import '../constants/constants'
+// import '../constants/constants'
 // require('../search')
 
-import { renderReact } from 'hypernova-react'
-import SearchMain from '../search'
+import search from '../search'
 
-export default renderReact('search', SearchMain)
+export default search
```
lauherkcambria: this change definitely makes this class cleaner for sure

lauherkcambria:
Also, @ratneshraval 5 spec tests are failing

app/views/facilities/index.html.erb
ratneshraval: I agree we shouldn't be doing pretender:true everywhere. This is just to make current pages work. Going forward we need to evaluate on case by case basis.

lauherkcambria: 👍

PR #87, opened 2017-06-14
RFA modeling

lauherkcambria:
- n/A

- #146343645

- we need to be able to have accurate models that represent the business objects in forms to make it easier to code and reuse throughout coding. 

This PR is intended to increase the amount of discussion and give us a solid ground for consensus on modeling. Once consensus is found, model testing is still needed. 

- no, but https://www.pivotaltracker.com/story/show/146466273 does depend on it's completion.

- n/A

- 100 -> 49.5

-0


app/models/alternative_name.rb
```diff
@@ -0,0 +1,4 @@
+class Name < CalsBase
+  attr_accessor :first, :middle, :last
```
ratneshraval: Class name change and add type.

app/models/cals_application.rb
```diff
@@ -0,0 +1,2 @@
+class CalsApplication < CalsBase
+end
```
ratneshraval: I think this file is irrelevant now, right?

app/models/license.rb
```diff
@@ -0,0 +1,11 @@
+class License < CalsBase
+  attr_accessor :agency_name, :license_type, :license_history_type
```
ratneshraval: should change the name to rfa_license

app/models/non_resident.rb
```diff
@@ -0,0 +1,4 @@
+class NonResident < CalsBase
+attr_accessor :legal_first, :legal_middle, :legal_last
+
```
ratneshraval: Need belongs_to?

ratneshraval:
Modeling looks good 👍. Please add tests.

PR #88, opened 2017-06-14
import css/images from woodduck

lavanyareddyreddy:
-n/a

-https://www.pivotaltracker.com/n/projects/1662141/stories/147229345

- import CSS files from cwds/reserch-design project, so that styling created by wood duck is consistent in CALS

-no

-no

-n/a

-n/a


lauherkcambria:
i think a lot of this needs to be under the vendor subdirectory instead of app/assets/stylesheets/vendor directory -- any other thoughts? also before approving this, i thought we were making an npm package with the shared css -- is that now not the case? @ratneshraval

lauherkcambria:
closing this PR as it is being replaced with the woodduck rails gem. Thank you for the effort and time you put in though @lavanyareddyreddy 😄 👍

PR #97, opened 2017-06-23
Rfa 01 a api dictionaries

lauherkcambria:
- N/A

- https://www.pivotaltracker.com/story/show/146466273

- this is the dictionary types for:

Name type
phone type
level of education
race/ethnicity
gender
state (if there is any such api), if not lets make a static method to support this
residence ownership type
language type

**TODO: testing of these api methods should be in the related controller tests, so no tests were added as a part of this PR.**

#### USAGE:

put in private method in controller: 

```
  def applicant_household_address_helper
    Helpers::RFA::ApplicantHouseholdAddress.new(auth_header: session['token'])
  end
```

use as follows:

 ` applicant_household_address_helper.residence_types`

- no

- no

- 98 -> 88

- no js diff


lauherkcambria:
retest please

lauherkcambria:
retest please

lauherkcambria:
Jenkins, retest this please.

lauherkcambria:
Jenkins, retest this please.

app/models/concerns/RFA/applicant_api_protocol_provider.rb
```diff
@@ -0,0 +1,26 @@
...
+      JSON.parse(response.body)
+    end
+
+    def gender_types(auth_header)
+      response = FaradayCals.get("/dictionaries/genders?token=null", auth_header)
+      JSON.parse(response.body)
+    end
+
+    def language_types(auth_header)
+      response = FaradayCals.get("/dictionaries/languages?token=null", auth_header)
```
ratneshraval: Let's use single quote where we don't need string interpolation.

app/models/concerns/RFA/applicant_api_protocol_provider.rb
```diff
@@ -0,0 +1,26 @@
+module Concerns::RFA::ApplicantApiProtocolProvider
+  extend ActiveSupport::Concern
+  include Concerns::BaseApiProtocolProvider
+
+  class_methods do
+    def education_levels(auth_header)
```
lavanyareddyreddy: How about something like this to DRY it?
```
 [: education_levels, : race_types, : gender_types, : language_types].each do |method_name|
     send :define_method, method_name do | auth_header| 
       json_response(get_dictionaries method_name.to_s.dasherize)
     end
  end

 private
   def get_dictionaries type
     response = FaradayCals.get("/dictionaries/#{type}", auth_header)
     JSON.parse(response.body)
   end
```

app/models/concerns/RFA/applicant_household_address_api_protocol_provider.rb
```diff
@@ -0,0 +1,11 @@
+module Concerns::RFA::ApplicantHouseholdAddressApiProtocolProvider
```
lavanyareddyreddy: This concern does not looks like something that will be used to deal with ApiProtocolProvider so isn't the name `ApplicantHouseholdAddressApiProtocolProvider` a bit misleading?

Cant it just be ApplicantHouseholdAddressDictionaries? 

And with autoloading concerns, like discussed, module name can just be `RFA::ApplicantHouseholdAddressDictionaries`

app/models/helpers/RFA/applicant.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def model_class
+    RFA::Applicant
+  end
+
+  def gender_types
+    RFA::Applicant.gender_types(auth_header)
+  end
+
+  def  race_types
```
lavanyareddyreddy: There is extra space

app/models/rfa/applicant_household_address.rb
```diff
@@ -1,6 +1,7 @@
 class RFA::ApplicantHouseholdAddress < Address
+  include Concerns::RFA::ApplicantHouseholdAddressApiProtocolProvider
   attr_accessor :property_type, :own_rent_lease,
-                :weapons, :body_of_water, :directions
+  :weapons, :body_of_water, :directions
```
lavanyareddyreddy: Can this be moved to line 3?

lavanyareddyreddy:
May be we can add tests on RFA applicant like below(replacing controller action calls with RFA applicant dictionary methods) and commit cassettes as well

```
describe RfaController do
  before(:each) do
    allow(controller).to receive_messages(:authenticate_with_cwds => true)
  end

  let(:expected_parsed_responses) do
    {residence_ownership_types: {'items' => [{'id' => 1, 'value' => 'Own'}, {'id' => 2, 'value' => 'Rent'}, {'id' => 3, 'value' => 'Lease'}]},
     languages: {'items' => [{'id' => 1, 'value' => 'Chinese Mandarin'}, {'id' => 2, 'value' => 'English'}, {'id' => 3, 'value' => 'French'},
                             {'id' => 4, 'value' => 'Russian'}, {'id' => 5, 'value' => 'Spanish'}]},
     name_types: {'items' => [{'id' => 1, 'value' => 'Alias'}, {'id' => 2, 'value' => 'Legal'}, {'id' => 3, 'value' => 'Maiden'}, {'id' => 4, 'value' => 'Preferred'},
                              {'id' => 5, 'value' => 'Other'}]},
     phone_number_types: {'items' => [{'id' => 1, 'value' => 'Cell'}, {'id' => 2, 'value' => 'Home'}, {'id' => 3, 'value' => 'Work'}]}
    }
  end

  [:residence_ownership_types, :languages, :name_types, :phone_number_types].each do |method_name|
    describe "GET #{method_name}" do
      it "renders #{method_name}"  do
        get method_name

        expect(JSON.parse response.body).to eq expected_parsed_responses[method_name]
      end
    end
  end
end
```

ratneshraval:
The dictionaries need to be models and follow our pattern of models and api provider. Hence they can all use our base methods like all and find_by_id; just like the Facility model.
That will also keep these dictionaries independent of any other models.

lavanyareddyreddy:
@ratneshraval Not sure why these dictionaries needs to be models, may be there can be a dictionaries model but it doesn't seem like a good idea to create models for different dictionaries because these dictionary concerns can be reused in multiple models.

Also, not sure what is the use case of something like find_by_id/find_by_name base methods because these are just supposed to be lists for our drop downs and there won't be any additional information apart from 'id' and 'value' of a specific dictionary that we may display by itself.

lauherkcambria:
Jenkins, retest this please.

ratneshraval:
I am ok with this implementation for now. I've added a model implementation so we can use that in future as need arises. For now we can move forward and work on other stories.

app/models/concerns/RFA/applicant_api_protocol_provider.rb
```diff
@@ -0,0 +1,26 @@
...
+      response = FaradayCals.get('/dictionaries/education-level-types?token=null', auth_header)
+      JSON.parse(response.body)
+    end
+
+    def race_types(auth_header)
+      response = FaradayCals.get('/dictionaries/ethnicity-types?token=null', auth_header)
```
ratneshraval: There are 2 different dictionaries. One for race types and one for ethnicity types.

PR #100, opened 2017-06-27
Rfa create react work

anuroop15:


ratneshraval:
Fixing the test in the parent branch

PR #107, opened 2017-06-29
Applicant save

ratneshraval:
- No tests added. 

- https://www.pivotaltracker.com/story/show/147277959

- Controller action to save applicant by calling api. 

- No.

-

- 78%

- 74


app/models/cals_base.rb
```diff
@@ -8,4 +8,8 @@ def self.class_name_downcase_pluralized
   def self.api_resource_path
     name.demodulize.downcase.pluralize
   end
+
+  def self.parent_path
```
lauherkcambria: what is parent path and what does it represent?

app/models/concerns/RFA/applicant_api_protocol_provider.rb
```diff
@@ -5,12 +5,10 @@ module Concerns::RFA::ApplicantApiProtocolProvider
   class_methods do
     def create(auth_header, parent_id, body)
       byebug
```
lauherkcambria: remove byebug

app/models/concerns/RFA/applicant_api_protocol_provider.rb
```diff
@@ -5,12 +5,10 @@ module Concerns::RFA::ApplicantApiProtocolProvider
...
-      response = FaradayCals.post("/rfa-1a-forms/#{parent_id}/#{api_resource_path}", auth_header, body)
+      response = FaradayCals.post("/#{parent_path}/#{parent_id}/#{api_resource_path}", auth_header, body)
```
lauherkcambria: personally, this feels like over engineering, and part of the reason why it 'feels' like this architecture is more complicated than it is

app/models/helpers/model_helper_base.rb
```diff
@@ -16,8 +16,12 @@ def all
     model_class.all(auth_header)
   end
 
+  # def create(url_path, body)
```
lauherkcambria: remove commented method

app/models/rfa/applicant.rb
```diff
@@ -14,32 +14,28 @@ class RFA::Applicant < CalsBase
...
-  # desired children should move to application
-
   # has many license_history
 
-  # def other_names=(val)
-  #   @other_names = JSON.parse(val).map { |itm| AlternativeName.new(itm) }
-  # end
+  def other_names=(val)
+    @other_names = val.map { |itm| AlternativeName.new(itm) } if val.present?
+  end
```
lauherkcambria: why have @ variable for other_names when class doesn't have other_names as an attr_accesor? what is the usage of these methods, and what is the passed in parameter 'val'? can it be more descriptive?

app/models/cals_base.rb
ratneshraval: Some resources need a parent path like applicant needs /application/:id, so this code is an attempt plan for that.

app/models/concerns/RFA/applicant_api_protocol_provider.rb
ratneshraval: Good catch. I'll remove it with some other updates.

app/models/helpers/model_helper_base.rb
ratneshraval: Can we leave this in place. This is my attempt to move the create to base class.

app/models/rfa/applicant.rb
ratneshraval: @ other_names was added to the model to support names being part of the applicant object.

lauherkcambria: in that case, i think :other_names needs to be added as an attr_accessor

ratneshraval: I think it is already there.

PR #109, opened 2017-06-30
Dictionaries update

ratneshraval:
- Tests will be added to related controller. 

- A different implementation of dictionary. Using Ethnicity and Race dictionary to show. 

- No 

- No

- Ruby 78%, JS same as before. 

- Same as before.


app/models/concerns/base_cals_api_protocol_provider.rb
```diff
@@ -4,12 +4,12 @@ module Concerns::BaseCalsApiProtocolProvider
...
   class_methods do
     def all(auth_header)
-      response = FaradayCals.get('/' + class_name_downcase_pluralized, auth_header)
+      response = FaradayCals.get('/' + api_resource_path, auth_header)
       JSON.parse(response.body)['items'].map { |itm| new(itm) }
     end
 
     def find_by_id(id, auth_header)
-      response = FaradayCals.get("/#{class_name_downcase_pluralized}/#{id}", auth_header)
+      response = FaradayCals.get("/#{api_resource_path}/#{id}", auth_header)
```
lavanyareddyreddy: Glad you replaced it with api_resource_path.

app/models/dictionary_base.rb
```diff
@@ -1,7 +1,11 @@
 class DictionaryBase < CalsBase
-  Concerns::BaseCalsApiProtocolProvider
```
lavanyareddyreddy: Great, we can also remove concerns namespace at some point by adding it to auto load

app/models/helpers/dictionaries/race_type_helper.rb
```diff
@@ -1,6 +0,0 @@
-class Helpers::Dictionaries::RaceTypeHelper < Helpers::ModelHelperBase
-
-  def model_class
```
lavanyareddyreddy: Great that you got rid of these!

app/models/dictionary_base.rb
ratneshraval: Can you create a story for that based on the article you found.

PR #110, opened 2017-07-05
Residence save

ratneshraval:
- Yes

- https://www.pivotaltracker.com/story/show/147278447
- https://www.pivotaltracker.com/story/show/148159255

- Residence card api save call. Address type dictionary added.. 

- No

- No.

- Ruby: 76.5%, JS: Same

- 116 -> 116


app/controllers/rfa/residence_controller.rb
```diff
@@ -0,0 +1,11 @@
+class Rfa::ResidenceController < CalsBaseController
+
+  def create
+    post_data = request.body.read
+    rfa_residence = rfa_residence_helper.create(params[:a01_id], post_data)
```
lauherkcambria: where is rfa_residence used? should this be an @ variable>?

app/models/concerns/RFA/applicant_api_protocol_provider.rb
```diff
@@ -4,7 +4,6 @@ module Concerns::RFA::ApplicantApiProtocolProvider
 
   class_methods do
     def create(auth_header, parent_id, body)
-      byebug
```
lauherkcambria: nice catch

app/models/concerns/RFA/application_residence_api_protocol_provider.rb
```diff
@@ -1,8 +1,14 @@
-module Concerns::RFA::ApplicantHouseholdAddressApiProtocolProvider
+module Concerns::RFA::ApplicationResidenceApiProtocolProvider
```
lauherkcambria: this is a better name change, nice

app/models/dictionaries/address_type.rb
```diff
@@ -0,0 +1,3 @@
+class Dictionaries::AddressType < DictionaryBase
```
lauherkcambria: so are we going to have a bunch of empty classes like this for all the dictionary types?

app/models/rfa/application_residence.rb
```diff
@@ -0,0 +1,32 @@
...
+    :residence_ownership, :weapon_in_home, :body_of_water_exist,
+    :body_of_water_description, :directions_to_home, :home_languages
+
+  attr_accessor :others_using_residence_as_mailing, :other_people_using_residence_as_mailing
+
+  #has_many :non_residents
+
+  #has_many :addresses
+
+  def addresses=(val)
```
lauherkcambria: lets stay away from parameters like val or string --- parameter should be named after what it is. also isn't this just a setter method...? in which case, we don't need since we have attr_accessors. (i.e. since we have attr_accessor :addresses, we automattically are given the addresses= method and addresses get method). what's the use case?

app/controllers/rfa/residence_controller.rb
ratneshraval: Should be @. I saw the code climate complaint about it but it is not used in UI yet, so the connecting react to rails story can convert this to @

app/models/dictionaries/address_type.rb
ratneshraval: Yes, unless we do further refactoring.

app/models/rfa/application_residence.rb
ratneshraval: This is custom setter hence 'val' naming. The default setter will only convert json to hash, this will covert it to address object.

app/models/concerns/RFA/application_residence_api_protocol_provider.rb
ratneshraval: 👍

app/controllers/rfa/residence_controller.rb
lauherkcambria: whats the redirect on this btw>??

lauherkcambria:
test cases to be addressed later

PR #118, opened 2017-07-11
Add Another Applicant

anuroop15:
Pre-int issue


app/javascript/forms/cardsMain.jsx
```diff
@@ -7,10 +7,14 @@ import Employment from './employment_card'
 export default class Cards extends React.Component {
   constructor (props) {
     super(props)
+    this.clickClose = this.clickClose.bind(this)
     this.state = {
       isFocused: {}
     }
   }
+  clickClose (id) {
```
ratneshraval: You'll probably have to rename this to properly reflect that this is for applicant card close. When we lift the state up, other close methods will probably end up here.

app/javascript/forms/cardsMain.jsx
```diff
@@ -22,30 +26,32 @@ export default class Cards extends React.Component {
   }
   render () {
     const {formData} = this.state
+    let showClose = (this.props.id === 0) ? false : true
     return (
       <div className='cards'>
-        <div id="nameSection" onClick={this.toggleOnFocus.bind(this, 'nameSection')} className={(this.state.isFocused['nameSection']) + ' ' + 'card name-section double-gap-top'}>
+        {showClose && <span onClick={() => this.clickClose(this.props.id)} className='pull-right glyphicon glyphicon-remove' />}
```
ratneshraval: I think this syntax is experimental and we should stick to bind method.

ratneshraval:
@lauherkcambria 
2. There is an X button to delete applicant card

lauherkcambria:
ready to go!

PR #124, opened 2017-07-15
add concerns to auto load paths

lavanyareddyreddy:
-no

- #148429363

- add concerns and helpers to auto load paths

- no

-no

- no

- no

### DIFF
--- a/app/controllers/concerns/authentication_provider.rb
+++ b/app/controllers/concerns/authentication_provider.rb
@@ -1,4 +1,3 @@
-
 module AuthenticationProvider
   extend ActiveSupport::Concern
 
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,7 +1,4 @@
-
-
 class SearchController < CalsBaseController
-
 
   def index
     @landingPage_url = SANDBOX_LANDING_URL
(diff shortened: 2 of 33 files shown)


ratneshraval:
I don't think we need this anymore since PR #131 fixes the problem without customized autoloading. 
If everyone agrees then I'll close this PR>

lavanyareddyreddy:
No, I dont agree. #131 does not fix the problem of all the unnecessary `Concern::` floating around in the entire codebase. 

Also, this story's acceptance criteria has nothing detailing any other problems apart from cleaning up the name spaces.

ratneshraval:
So cleaning up namespacing will still work without autoloading.
If this allows us to remove Concerns:: from namespace and still lets us modify code without restarting rails server then I say we go ahead with this. If not, what purpose will it serve?

lavanyareddyreddy:
Modify code without restarting rails server has nothing to do with this PR. 

I don't think autoload is necessary because looks like now `app/*/concern` is included in autoload paths by default so, I will take that out.

ratneshraval:
Checked with @lavanyareddyreddy and we agreed to close this PR. So closing this as PR #131 was able to handle the change without coding for autoload at start.

PR #130, opened 2017-07-19
Residence refactor

ratneshraval:
- All existing tests updated to pass

- https://www.pivotaltracker.com/story/show/148091485 (WIP)
- https://www.pivotaltracker.com/story/show/148679103
- https://www.pivotaltracker.com/story/show/148766707

- Address card refactor to bring state up to index. Some other bug fixes


app/javascript/forms/addressCard.js
```diff
@@ -1,112 +1,104 @@
...
+  city: '',
+  state: {
+    id: '',
+    value: ''
+  },
+  type: {
+    id: '2',
+    value: 'Mailing'
+  }
+})
```
lauherkcambria: there has got to be a better way to do this -- why not pass in the  type value via a method? i.e. ```
defaultFields(type) {
return {
...,
type:
{
id:'2',
 value: type
}
}
} 
```

Comment (lauherkcambria) [inline app/javascript/forms/addressCard.js]: also, this breaks the rule of one component doing 1 thing

File: app/javascript/constants/constants.js [hunk 2]
```diff
@@ -63,11 +63,11 @@ const CountyList = Object.freeze({
 const yesNo = Object.freeze({
   'items': [
     {
-      'id': 1,
+      'id': true,
       'value': 'Yes'
     },
     {
-      'id': 2,
+      'id': false,
```
lauherkcambria: why? usually with hashes id's aren't booleans

app/javascript/forms/addressCard.js
ratneshraval: I agree. I don't like how the address isn't in its own component, which is resulting in complicated address update code. 
I added a new story to refactor and move address to it's own component. https://www.pivotaltracker.com/story/show/149177125

app/javascript/constants/constants.js
ratneshraval: Yes/no dropdowns require us to send boolean value to API, so in this case the ID is boolean and text is yes/no.

app/javascript/forms/addressCard.js
lauherkcambria: nice 👍

app/javascript/constants/constants.js
lauherkcambria: ah, gotcha. good to know! 👍

PR #132, opened 2017-07-19
name Card and applicant Card Group testcases

anuroop15:
-yes

-NA

- Added tests to check click functionality

-NA

-npm run karma

-JS -> 63

-NA


test/javascript/forms/applicantCardsGroup_tests.js
```diff
@@ -0,0 +1,29 @@
...
+                                              educationLevels={educationLevels.items}
+                                              genderTypes={genderTypes.items}
+                                              raceTypes={ethnicityTypes.items}
+                                              languageTypes={languageTypes.items}
+                                              addCard={isApplicantAdded.items} />)
+  const applicantGroup = applicantRender.getRenderOutput()
+
+  it('Spy on Button Click doesnot call', () => {
+    //const addCardBut = applicantRender
+    const addCardBtn = applicantGroup.props.children[1].props.children.props.children
```
ratneshraval: Can you find it by id or some other means? If not, enzyme seems to be better for us.

test/javascript/forms/nameCard_tests.js
```diff
@@ -0,0 +1,71 @@
...
+    nameField, otherName, renderedCard,
+    renderedDOM, renderedDom
+  beforeEach (() => {
+    setCardState = sinon.spy()
+    isNameCardRemoved = sinon.spy()
+    handleNameChangeFun = sinon.spy()
+    nameType = nameTypes
+    nameField = {
+      first_name: '',
+      middle_name: '',
```
ratneshraval: I was able to bring my blankObject into phone card tests by exporting it. Something to keep in mind, so you don't have to keep it up to date in 2 places.

PR #135, opened 2017-07-25
Phone card tests

ratneshraval:
- Yes

- Adding more phone card tests

#### Code coverage
- JS 55%


ratneshraval:
Ready for review. phoneNumberCardsGroup.jsx is at 100%
phoneNumberFields.js at 71%

test/javascript/helpers/constants.js
```diff
@@ -317,7 +317,7 @@ const ethnicityTypes = Object.freeze({
   ]}
 )
 
-const phoneTypes = Object.freeze(
+export const phoneTypes = Object.freeze(
```
ratneshraval: The article doesn't mention that one is better than other way. But for readability, having export right next to the class/const is better than scrolling down and trying to figure out if it is exported.

PR #137, opened 2017-07-28
Rfa directory structure

ratneshraval:
- Yes all tests pass

- Restructuring for index/list view story

- 'forms' directory updated to reflect 'rfa_forms/rfa01a_edit'

- 54% JS

- 86


ratneshraval:
> i'm not really sure what this refactor gives us that we didn't have before -- as far as granularity is concerned, i don't know if the renaming to be 'edit' really makes sense as this UI isn't just for editing the form but also used for new/create. either way works for me though if this change makes more sense to everyone else then great lets go for it 👍 😄

Technically it is edit for existing application but we can rename it to something else which makes more sense. What would be your suggestion on the naming for that?

lavanyareddyreddy:
codeclimate -> 70 issues to fix. are we ignoring them?

ratneshraval:
> codeclimate -> 70 issues to fix. are we ignoring them?

They are all "similar code found" errors and we can add a refactor story to look at them later but safe to ignore now.

lauherkcambria:
> Technically it is edit for existing application but we can rename it to something else which makes more sense. What would be your suggestion on the naming for that?

For me the way it was already named made sense to me and i don't really see the need for this change. But as I was saying before, if this makes more sense this way for the rest of the team, lets go for it. 😄

PR #142, opened 2017-08-04
Other adults connect and save api

lauherkcambria:
-

-https://www.pivotaltracker.com/story/show/147282229

- connect api to other adults card section for save. write methods that will be used in future. clean up concerns and helpers.

-

-

-

-


app/controllers/cals_base_controller.rb
```diff
@@ -7,6 +7,7 @@ class CalsBaseController < ApplicationController
   protected
 
   def authenticate_with_cwds
+session.delete(:token) if params[:token].present?
```
ratneshraval: Formatting.

app/models/concerns/rfa_base_api_protocol_provider.rb
```diff
@@ -0,0 +1,16 @@
+module Concerns::RfaBaseApiProtocolProvider
```
ratneshraval: Nice.

app/javascript/rfa_forms/rfa01a_edit_view/OtherAdultsCardsGroup.js
```diff
@@ -56,12 +67,32 @@ export default class OtherAdultsCardsGroup extends React.Component {
...
+    let relationshipList = Immutable.fromJS(otherAdults[index].relationship_to_applicants[0])
+
+    relationshipList = relationshipList.setIn(['relationship_to_applicant', 'id'], value)
+    relationshipList = relationshipList.setIn(['relationship_to_applicant', 'value'], this.props.relationship_types[value].value)
+
+    otherAdultsList = otherAdultsList.update(index, x => x.set('relationship_to_applicants', relationshipList))
+    this.props.setParentState('otherAdults', otherAdultsList.toJS())
+  }
+  handleToWhom (index, value) {
+    let otherAdults = checkArrayObjectPresence(this.props.otherAdults) || [otherAdultsDefaults]
```
ratneshraval: Let's whiteboard this toWhom section. It needs to be built at index level for other components to use.

lauherkcambria:
>  Code climate is complaining about missing gemfile.lock
> Can you add it back?

lol. of course it would. for sure will do that now

lauherkcambria:
> Let's whiteboard this toWhom section. It needs to be built at index level for other components to  > use.

i was thinking about it, and think it would be a good idea for us to pair on it and pursue the solution we were discussing yesterday. That said though, since those methods aren't attached to anything it won't prevent/delay merging .

lauherkcambria:
code climate is saying we need to whitelist our params....weird....hahah

app/javascript/rfa_forms/rfa01a_edit_view/OtherAdultsCardsGroup.js
```diff
@@ -2,21 +2,31 @@ import Immutable from 'immutable'
...
+  //   'value': ''
+  // },
+  'first_name': '',
+  'middle_name': '',
+  'last_name': '',
+  // 'name_suffix': {
+  //   'id': 0,
+  //   'value': ''
+  // },
+  'date_of_birth': '',
```
lavanyareddyreddy: @ratneshraval , @lauherkcambria : like nesh said do you want to remove date_of_birth default.

app/controllers/rfa/a01_controller.rb
```diff
@@ -26,7 +26,7 @@ def edit
     # @race_types = dictionaries_helper.race_types
     @address_types = dictionaries_helper.address_types
     @relationship_types = dictionaries_helper.relationship_types
-
+    @relationship_to_applicant_types = dictionaries_helper.relationship_to_applicant_types
```
lavanyareddyreddy: @lauherkcambria : we might get conflicts on this dictionary :(

app/javascript/rfa_forms/rfa01a_edit_view/OtherAdultsCardsGroup.js
lauherkcambria: look at the controller, i'm doing an except there

lavanyareddyreddy: Yes, but eventually we will need to send dob in future. the code in controller is only for time being. like nesh said we dont want to send blank dob if the user doesn't enter any value. @ratneshraval : if it makes sense  to you guys I am fine to leave it as it is,  but wanted to make sure as, dob defaults is removed from minorchildren defaults as well.

PR #144, opened 2017-08-07
remove link to facilities and search routes

lauherkcambria:
- N/A

- https://www.pivotaltracker.com/story/show/149970760

- remove index link to facilities and search, remove routes, and comment out tests to ready for push to integration environment


ratneshraval:
Looking at this, it feels like it is probably good idea to just hide the link and leave routes/tests in place.
What do you guys/gal think?

lauherkcambria:
I am ok with that 👍

lavanyareddyreddy:
I agree with @ratneshraval

PR #145, opened 2017-08-07
auto populate to whom

lauherkcambria:
-N/A

- https://www.pivotaltracker.com/story/show/148788567

- AUTO UPDATE to whom fields based off of user input in applicant

-no

- rake tmp:cache:clear; rails assets:precompile.


lauherkcambria:
we will make a pivotal story to address the posting of the to whom data and get this in now as there are some api contract discrepencies on the API side that we will wait on completion for in order to minimize duplicate work.

ratneshraval:
Looks like you extracted setToWhom but still passing it down as props. You can just import it in minorCardFields and not have to worry about passing as props.

lauherkcambria:
good catch @ratneshraval, i'll push a fix for that. thanks for fixing those tests @lavanyareddyreddy!👍 👏 😄

ratneshraval:
This can use few extra test cases, testing setToWhom function to increase coverage.

PR #147, opened 2017-08-07
fix child adopted field

lavanyareddyreddy:
no
-#147206857 

- fix child adopted field to keep it consistent with the design

-no

- no

-n/a

-n/a


lauherkcambria:
image

should the second column of 'To Whom' and 'do you financially support this child' line up?

lauherkcambria:
image
changed to this

PR #151, opened 2017-08-08
Update component with props.

anuroop15:
- yes

-#149593933

- This story allows users to navigate from application list page to edit page with concerned props

-NA

- npm run dev

- 65%

- 69


app/javascript/rfa_forms/rfa01a_edit_view/index.js
```diff
@@ -15,9 +15,10 @@ export default class Rfa01EditView extends React.Component {
     super(props)
     this.state = {
       focusComponentName: '',
+      //application: this.props.application
       application: {
-        applicants: [],
-        residence: {},
+        applicants: (this.props.application.applicants && this.props.application.applicants.items.length > 0) ? this.props.application.applicants.items : [],
```
ratneshraval: I'll look into this and the phone number map code change but this looks good for now.

PR #156, opened 2017-08-18
WIP - New name card for suffixs and Prefixs

anuroop15:
- yes

-#150017725 & #150017731

-This story creates user to add Prefix and suffix to their names

- NA

- npm run dev

- 63%

- 94


app/models/concerns/rfa/application_api_protocol_provider.rb
```diff
@@ -13,5 +13,22 @@ def create_application(auth_header)
...
+      JSON.parse(response.body)
+    end
+    def name_suffix_types(auth_header)
+      response = FaradayCals.get('/dictionaries/name-suffix-types', auth_header)
+      JSON.parse(response.body)
+    end
+    def name_prefix_types(auth_header)
+      response = FaradayCals.get('/dictionaries/name-prefix-types', auth_header)
+      JSON.parse(response.body)
+    end
```
lauherkcambria: These do not need to be here and should follow the architecture we have for all other dictionaries.

PR #157, opened 2017-08-18
Data load residence card

lavanyareddyreddy:
- yes, yes

-#150111746

-Validate data load for Residence card
added tests for Address card, About the residence, Residence Main.

- no

- npm run karma, ./bin/server

-rfa_forms/rfa01a_edit_view/  75.22%

- n/a


app/javascript/rfa_forms/rfa01a_edit_view/addressCard.js
```diff
@@ -45,50 +44,62 @@ export default class AddressCard extends React.Component {
     data = data.update(index, x => x.set(key, value))
     this.props.setParentState('addresses', data.toJS())
   }
-
   render () {
-    const hiddenMailingSameAsPhysical = this.props.physicalMailingSimilar === 'false' ? '' : 'hidden'
+    const physicalAddressFields = (this.props.addresses !== undefined && this.props.addresses[0]) ? this.props.addresses[0] : blankPhysicalAddress
```
ratneshraval: This needs to find by type and not find by index. What happens when there isn't a physical address typed and only mailing is entered. Wouldn't index 0 will have mailing address?

app/javascript/rfa_forms/rfa01a_edit_view/addressCard.js
```diff
@@ -45,50 +44,62 @@ export default class AddressCard extends React.Component {
...
+    const mailingAddressFields = (this.props.addresses !== undefined && this.props.addresses[1]) ? this.props.addresses[1] : blankMailingAddress
+    const mailingAddress = this.props.physicalMailingSimilar !== undefined ? this.props.physicalMailingSimilar : ''
```
ratneshraval: Sometime a combo of default props and conditional rendering will make these if blocks go away :)

lauherkcambria: great point @ratneshraval! @lavanyareddyreddy, if we have enough time in this sprint, do you think this can be addressed on this PR? If not, I think we should make note of this change for now and plan to handle it later as a refactor. what do you two think? 😄

lavanyareddyreddy: Yeah great idea, we can get this one in instead so that I don't have to deal with merge conflicts on a long standing PR. I will make a note of it.

lauherkcambria: sounds good, i'll add my approval so we can get it merged.

app/javascript/rfa_forms/rfa01a_edit_view/addressCard.js
lauherkcambria: @lavanyareddyreddy does this need to be addressed before merge?

lavanyareddyreddy: @lauherkcambria, @ratneshraval  : yes it is fixed, its now based on mailing and physical address type based assignment. please have a look.

PR #158, opened 2017-08-21
data load for foster care

lavanyareddyreddy:
- no

- #150111756

- Validate data load for foster history card

- no

- ./bin/server

- n/a

- n/a


app/javascript/rfa_forms/rfa01a_edit_view/FosterCareHistoryFields.js
```diff
@@ -162,13 +162,14 @@ export class FosterCareHistoryFields extends React.Component {
...
-    const hiddenQ3 = q3History.was_previously_licensed === 'true' ? '' : 'hidden'
-    const hiddenQ4 = q4History.was_employed_or_volunteered === 'true' ? '' : 'hidden'
-    const hiddenQ5 = q5History.had_denials === 'true' ? '' : 'hidden'
-    const hiddenQ6 = q6History.had_suspensions_revocations === 'true' ? '' : 'hidden'
+    const hiddenQ1 = (q1History.was_previously_licensed === 'true' || q1History.was_previously_licensed === true) ? '' : 'hidden'
+    const hiddenQ2 = (q2History.have_applied_for_adoption === 'true' || q2History.have_applied_for_adoption === true) ? '' : 'hidden'
+    const hiddenQ3 = (q3History.was_previously_licensed === 'true' || q3History.was_previously_licensed === true) ? '' : 'hidden'
+    const hiddenQ4 = (q4History.was_employed_or_volunteered === 'true' || q4History.was_employed_or_volunteered === true) ? '' : 'hidden'
+    const hiddenQ5 = (q5History.had_denials === 'true' || q5History.had_denials === true) ? '' : 'hidden'
+    const hiddenQ6 = (q6History.had_suspensions_revocations === 'true' || q6History.had_suspensions_revocations === true) ? '' : 'hidden'
```
lauherkcambria: nice! 😄 there are a couple ways around this that we may want to explore as well. for example, instead of adding the || statement, we can do JSON.parse(q1History.was_previously_licensed). This will always return a boolean. Another solution would be to use == rather than === (=== also includes type in equality comparison, where as == does not). Another possible thing we could do to clean up code even more would be to make a stringToBool(string) {} method in the card helper. what Is nice with this approach of making a stringToBool method is that we could reuse it in a generic isFieldHidden(fieldValue) {} also in the cardHelper, and remove all of these lines in one fell swoop! 😄

lauherkcambria: we can do this as a refactor.

lavanyareddyreddy: Sure we can refactor later.

app/javascript/rfa_forms/rfa01a_edit_view/FosterCareHistoryFields.js
```diff
@@ -162,13 +162,14 @@ export class FosterCareHistoryFields extends React.Component {
...
     const q6History = fosterCareHistory.suspension_revocation_history_q6
+    const q7History = fosterCareHistory.was_subject_for_exclusion_order_q7
 
-    const hiddenQ1 = q1History.was_previously_licensed === 'true' ? '' : 'hidden'
-    const hiddenQ2 = q2History.have_applied_for_adoption === 'true' ? '' : 'hidden'
-    const hiddenQ3 = q3History.was_previously_licensed === 'true' ? '' : 'hidden'
-    const hiddenQ4 = q4History.was_employed_or_volunteered === 'true' ? '' : 'hidden'
-    const hiddenQ5 = q5History.had_denials === 'true' ? '' : 'hidden'
-    const hiddenQ6 = q6History.had_suspensions_revocations === 'true' ? '' : 'hidden'
+    const hiddenQ1 = (q1History.was_previously_licensed === 'true' || q1History.was_previously_licensed === true) ? '' : 'hidden'
```
lavanyareddyreddy: Yeah as discussed below on line 172, plan is to refactor later

anuroop15: ok.

app/javascript/rfa_forms/rfa01a_edit_view/FosterCareHistoryFields.js
lavanyareddyreddy: @anuroop15 : sound good. but I am not using .map anywhere on that line.if possible, please point me to exact line.

PR #159, opened 2017-08-21
handle to whom method, validate other adults & minor children

lauherkcambria:
- added other adults tests

- https://www.pivotaltracker.com/story/show/150111754
- https://www.pivotaltracker.com/story/show/150111757

- validate all fields save and load for other adults and minor children fields 

- no

- rake tmp:cache:clear;rails assets:precompile

- 70%

- 68


app/controllers/rfa/a01_controller.rb
```diff
@@ -37,103 +35,105 @@ def applicant_params
...
-      return result
-    end
+  def adoption_history_params
+    params.require(:fosterCareHistory).permit(:id, :was_subject_for_exclusion_order_q7,
+                                              foster_care_licenses_q1: [:was_previously_licensed, agencies: [:name, type: %i[id value]]],
+                                              applications_for_adoption_q2: [:have_applied_for_adoption, facilities: []],
+                                              facility_operation_licenses_q3: [:was_previously_licensed, agencies: [:name, type: %i[id value]]],
+                                              employment_in_facilities_q4: [:was_employed_or_volunteered, facilities: []],
+                                              denial_history_q5: [:had_denials, agencies: [:name, type: %i[id value]]],
+                                              suspension_revocation_history_q6: [:had_suspensions_revocations, agencies: [:name, type: %i[id value]]])
```
lavanyareddyreddy: Indentation looks much better, wonder if there is a better way to manage this huge list of params

lauherkcambria: i agree on the indentation -- lets make a research chore to find out if there is a better way to manage this list.

anuroop15:
screen shot 2017-08-21 at 2 04 43 pm
@lauherkcambria : the toWhom dropDown was pre-selected. As per design this shouldn't be pre-selected When I have applicant information in name card.

anuroop15:
screen shot 2017-08-21 at 2 55 58 pm

PUT method was failing initially, as the request Payload was having 'appliantId' set to null. This will  fail as API doesn't accept this json

PR #163, opened 2017-08-27
UI for relation between applicants

lavanyareddyreddy:
### What tests were added (if relevant), and do all tests (rspec, karma, and integration CALS tests) pass?
-yes

- #147207477, #147282949

-Relationship between applicant - User Interface RFA 01A
-Relationship between applicants  - Connect API 

-no

-npm run karma, ./bin/server

-81.69%

-n/a


app/javascript/rfa_forms/rfa01a_edit_view/relationshipBetweenApplicantsCard.jsx
```diff
@@ -0,0 +1,63 @@
+
+import React from 'react'
+import {yesNo} from 'constants/constants'
```
anuroop15: Remove if not needed

app/javascript/rfa_forms/rfa01a_edit_view/relationshipBetweenApplicantsFields.js
```diff
@@ -0,0 +1,70 @@
...
+    const applicant1FirstName =  this.props.applicants[0] !== undefined && this.props.applicants[0].first_name ? this.props.applicants[0].first_name : false
+    const applicant2FirstName =  this.props.applicants[1] !== undefined && this.props.applicants[1].first_name ? this.props.applicants[1].first_name : false
+    return (
+      <form>
+        <div className='row'>
+          <DropDownField gridClassName='col-md-12' id='relationship_type'
+            selectClassName='reusable-select'
+            value={getDictionaryId(relationship.relationship_type)}
+            optionList={this.props.relationshipTypes}
+            label={'Relationship between ' + applicant1FirstName + ' and ' + applicant2FirstName}
```
anuroop15: This was showing only first name. Should it display first name and last name? We should check this with Business.

app/javascript/rfa_forms/rfa01a_edit_view/relationshipBetweenApplicantsCard.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+export default class RelationshipBetweenApplicantsCard extends React.Component {
+
+  constructor (props) {
+    super(props)
+    this.setState = this.setState.bind(this)
+    this.getFocusClassName = this.getFocusClassName.bind(this)
+  }
+
+  getFocusClassName (componentName) {
```
lauherkcambria: this method has been abstracted out to the cardhelper. let's remove it and use 

`import {getFocusClassName} from 'helpers/cardsHelper.jsx'`

app/javascript/rfa_forms/rfa01a_edit_view/relationshipBetweenApplicantsCard.jsx
```diff
@@ -0,0 +1,60 @@
...
+    return this.props.focusComponentName === componentName ? 'edit' : 'show'
+  }
+
+  setState (key, value) {
+    let newData = Immutable.fromJS(this.props.relationshipBetweenApplicants || blankValues)
+    newData = newData.set(key, value)
+    this.props.setParentState('relationshipBetweenApplicants', newData.toJS())
```
lauherkcambria: do we need this method? since we always have data with the `let data = this.props.relationshipBetweenApplicants || blankValues` won't 'new data' always have the correct value? if so,  we may be able to just pass through this.props.setParentState. 😄

app/javascript/rfa_forms/rfa01a_edit_view/relationshipBetweenApplicantsFields.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react'
+import Immutable from 'immutable'
+import {InputComponent} from 'components/common/inputFields'
+import {TextAreaComponent} from 'components/common/textArea'
+import {DropDownField} from 'components/common/dropDownField'
+import {dictionaryNilSelect, getDictionaryId} from 'helpers/commonHelper.jsx'
+
+export default class RelationshipBetweenApplicantsFields extends React.Component {
```
lauherkcambria: lets remove these padded lines

app/models/concerns/RFA/applicants_relation_api_protocol_provider.rb
```diff
@@ -0,0 +1,17 @@
...
+  extend ActiveSupport::Concern
+  include Concerns::RfaBaseApiProtocolProvider
+
+  class_methods do
+    def create(auth_header, parent_id, body)
+      response = FaradayCals.put("/#{parent_path}/#{parent_id}/#{api_resource_path}", auth_header, body)
+      new(JSON.parse(response.body))
+    end
+
+    def update(auth_header, parent_id, id, body)
```
lauherkcambria: lets remove the unused method argument id

app/models/helpers/dictionary.rb
```diff
@@ -54,22 +54,22 @@ def language_types
...
-    dictionaries[:address_types]=address_types
-    dictionaries[:relationship_types]= relationship_types
-    dictionaries[:license_types]= license_types
-    dictionaries[:relationship_to_applicant_types]= relationship_to_applicant_types
+    dictionaries[:residence_types] = residence_ownership_types
+    dictionaries[:ethnicity_types] = ethnicity_types
+    dictionaries[:address_types] = address_types
+    dictionaries[:relationship_types] = relationship_types
+    dictionaries[:license_types] = license_types
+    dictionaries[:relationship_to_applicant_types] = relationship_to_applicant_types
```
lauherkcambria: thanks for fixing these

app/models/rfa/applicant.rb
```diff
@@ -3,7 +3,7 @@ class Rfa::Applicant < CalsBase
 
   attr_accessor :language, :highest_education_level, :email, :date_of_birth, :ethnicity, :race, :gender,
     :driver_license_number,:driver_license_state, :email_address, :employment, :exclusion_order,
-    :first_name, :middle_name, :last_name, :name_suffix, :name_prefix
+    :first_name, :middle_name, :last_name, :name_suffix, :name_prefix, :validation_details
```
lauherkcambria: will `:validation_details` be in all models?

test/javascript/rfa_forms/rfa01a_edit_view/relationBetweenapplicantCard_tests.js
```diff
@@ -0,0 +1,60 @@
...
+      let relationShipField = componentMount.find('#RelationshipBetweenApplicantsCardSection')
+      relationShipField.simulate('click')
+      expect(setFocusStateSpy).toHaveBeenCalledWith('RelationshipBetweenApplicantsCard')
+    })
+
+  })
+
+})
+
+
```
lauherkcambria: nice tests! do you think we should also eventually add a test for the disappearing aspect of this card if we add/remove extra applicant ?

app/models/rfa/applicant.rb
lavanyareddyreddy: I had error related to :validation_details not being found in applicant model and residence model so I fixed it by adding the attribute, not sure if we need it in other models. we can fix with other models if we notice an error.

test/javascript/rfa_forms/rfa01a_edit_view/relationBetweenapplicantCard_tests.js
lavanyareddyreddy: Sure maybe. I will see if I can update the tests.

app/models/concerns/RFA/applicants_relation_api_protocol_provider.rb
lavanyareddyreddy: id is needed to fix an issue where applicant 1's name was being updated on applicant 2's section. but I don't see that issue any more even after removing id. So, I will remove Id and commit. and If we face any issue with applicant name not being updated we add this 'id' argument back.

app/models/rfa/applicant.rb
lauherkcambria: sounds good! thanks for looking into it and getting this info! 👍

app/javascript/rfa_forms/rfa01a_edit_view/relationshipBetweenApplicantsFields.js
lavanyareddyreddy: Removed the unwanted spaces.

PR #168, opened 2017-09-01
fix phone no null default and weird preference switching bug 

lauherkcambria:
- n/a

-https://www.pivotaltracker.com/story/show/149531637
https://www.pivotaltracker.com/story/show/149531635

- fix phone no null default and weird preference switching bug where it would switch to applicant 1

-no

-no

-n/a

-n/a


app/javascript/components/common/dropDownField.js
```diff
@@ -1,11 +1,12 @@
...
 import PropTypes from 'prop-types'
 
-const DropDownField = ({gridClassName, id, value, disable, selectClassName, optionList, onChange, label}) => (
+const DropDownField = ({gridClassName, id, value, disable, disableNullVal, selectClassName, optionList, onChange, label}) => (
   <div className={gridClassName}>
     <label>{label}</label>
     <select value={value} id={id} disabled={disable} onChange={onChange} className={selectClassName}>
-      {!disable && <option key='' value='' />} }
+      {!disable}
+      { !disableNullVal && <option key='' value='' />}
```
ratneshraval: the original code with !disable && didn't work?

app/javascript/rfa_forms/rfa01a_edit_view/applicantCard.jsx
```diff
@@ -75,6 +75,7 @@ export default class ApplicantCard extends React.Component {
           </div>
           <PhoneComponent
             phoneTypes={this.props.phoneTypes}
+            applicant_id={this.props.index}
```
ratneshraval: This is good for now but another solution for this is coming through the validation story where we had to use 'id prefix' to differentiate ids.

lauherkcambria: sounds good, but i'm not sure that would work -- the way it is in the last name validation PR right now uses index as the identifier which would be the same for applicant 1 and 2 causing the phone ids to be matching -- the phone id needs a unique identifier, is there a way to do that that without the applicant id?

PR #169, opened 2017-09-01
Fix date time to show mm/dd/yyyy

lauherkcambria:
- tests will be added upon switching to cleave.js

- https://www.pivotaltracker.com/story/show/150495530

- dob is in view like mm/dd/yyyy

- no

- no

- 75%


app/javascript/helpers/commonHelper.jsx
```diff
@@ -14,3 +14,24 @@ export const removeLegalNameType = (nameTypes) => {
   delete nameTypes[2]
   return nameTypes
 }
+
+export const FormatDoBForPersistance = (dateString) => {
```
ratneshraval: Isn't there any react date library we can use to switch formats, instead of doing it ourselves?

lauherkcambria: reason why we have to do it ourselves is because of the specific format the DB requires us to have -- i was unable to find a publicly available library that had this functionality.

ratneshraval: We have npm package called 'moment'. Take a look at that, it allows format and parse on various formats.

lauherkcambria: yeah, i spent more time trying to get moment to work with saving to the db than it took to just write the method, butthat said how would you recommend it's usage here? `moment(datestring, 'MM/DD/YYYY')` doesn't seem to work.

ratneshraval: try this
parsing: moment(datestring, 'MM/DD/YYYY')
formatting existing date object: moment().format('YYYY-MM-DD')

ratneshraval: export const FormatDoBForPersistance = (dateString) => {
  let m = moment(dateString, 'MM/DD/YYYY')
  return m.format('YYYY-MM-DD')
}

export const FormateDobForDisplay = (dateString) => {
  let m = moment(dateString, 'YYYY-MM-DD')
  return m.format('MM/DD/YYYY')
}

lauherkcambria: since this doesn't work we decided to go with original solution. 👍

PR #170, opened 2017-09-07
abstract dob logic and add to all places

lauherkcambria:
- na

- https://www.pivotaltracker.com/story/show/150495530

- replaces all dob fields with dob logic 

-

-

-

-


app/javascript/rfa_forms/rfa01a_edit_view/OtherAdultsCardsGroup.js
```diff
@@ -16,6 +16,7 @@ const otherAdultsDefaults = Object.freeze({
   'first_name': '',
   'middle_name': '',
   'last_name': '',
+  date_of_birth: '',
```
ratneshraval: missing quotes?

lauherkcambria: nice catch -- funny thing is that it totally works without quotes too haha 😄

lauherkcambria: actually i'll take out all the quotes haha

ratneshraval:
1. I like how DateOfBirth field is now a component, easy to replace maskedInput with Cleave when ready. Although, shouldn't it be named dateField so it can be used when there is a date field but not dateOfBirth?
2. Applicant history component somehow made it into this PR.

lauherkcambria:
TODO: add datepicker in everywhere else :{

lauherkcambria:
retest

app/javascript/components/common/DateFields.jsx
```diff
@@ -0,0 +1,117 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import FormField from './formField'
+import MaskedInputField from 'components/common/maskedInputField.jsx'
```
ratneshraval: Can you remove masked input from here?

app/javascript/components/common/OtherAdultsCardField.js
```diff
@@ -1,13 +1,29 @@
...
+const dateValidator = [{rule: 'isValidDate', message: 'date is invalid'}]
 
 export class OtherAdultsCardField extends React.Component {
+
+  constructor (props) {
+    super(props)
+
+    this.props.validator.addFieldValidation(this.props.idPrefix + 'date_of_birth', dateValidator)
+  }
+  validateOnBlur (value, type) {
```
ratneshraval: Please use validateOnBlur from validationHelper.jsx since you'll need forceUpdate()

app/javascript/helpers/validator.js
```diff
@@ -38,7 +39,7 @@ export default class Validator {
     return Immutable.Set(this.errors[fieldName] || [])
   }
 
-  isRequired (value) { // TODO: Rename to isNotEmpty
+  isNotEmpty (value) {
```
ratneshraval: I know I had a TODO here but this may break the nameCard validation, unless you changed there too. If not, can you revert this?

app/javascript/rfa_forms/rfa01a_edit_view/aboutApplicantCard.jsx
```diff
@@ -2,61 +2,83 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import {InputComponent} from 'components/common/inputFields'
 import {DropDownField} from 'components/common/dropDownField'
-import MaskedInputField from 'components/common/maskedInputField.jsx'
+import {DateOfBirthField} from './../../components/common/DateFields.jsx'
```
ratneshraval: Can we use top down reference here?

test/javascript/rfa_forms/rfa01a_edit_view/otherAdultsCardsGroup_tests.js
```diff
@@ -56,11 +58,7 @@ describe('Verify other adults Component View', function () {
   })
 
   it('verify date of birth', () => {
-    componentMount.update()
-    spyOn(componentMount.instance(), 'onFieldChange').and.callThrough()
-    let relationShipField = componentMount.find('#dateOfBirth')
-    relationShipField.simulate('change', {target: {value: '2017-01-01'}})
-    expect(setParentStateSpy).toHaveBeenCalledWith('otherAdults', [OtherAdultsCard])
+  // TODO will update when switching from react-maskedinput to cleave.js
```
ratneshraval: Since you switched to Cleave, can you add this test?

app/javascript/rfa_forms/rfa01a_edit_view/minorCardField.js
```diff
@@ -1,14 +1,28 @@
...
+    this.props.validator.addFieldValidation(this.props.idPrefix + 'date_of_birth', dateValidator)
+  }
+  validateOnBlur (value, type) {
+    this.props.validator.validateField(type, value)
+    this.forceUpdate()
+  }
+
   render () {
     const minor = this.props.minorChild
+    const mindorRuleId = this.props.idPrefix + 'date_of_birth'
```
anuroop15: better to have readable variable. I think it's a typo. if minorRuleId was what you have though of?, changing this will be good

app/javascript/components/common/OtherAdultsCardField.js
anuroop15: it would be good if we minimal to use forceUpdate(). If we relay on props and state, we doesn't need to have forceUpdate()
https://facebook.github.io/react/docs/react-component.html#forceupdate

test/javascript/rfa_forms/rfa01a_edit_view/otherAdultsCardsGroup_tests.js
lauherkcambria: karma still doesn't work with cleave still :(

PR #173, opened 2017-09-13
hotfix for weird on focus safari bug in employment income field

lauherkcambria:
hotfix for employment income bug

https://www.pivotaltracker.com/story/show/151067285


app/javascript/rfa_forms/rfa01a_edit_view/employmentCard.jsx
```diff
@@ -50,14 +52,22 @@ export default class Employment extends React.Component {
               label='Occupation' placeholder=''
               type='text' onChange={(event) => this.onEmploymentChange('occupation', event.target.value)} />
 
-            <div className='col-md-3'>
+            <div className='col-md-3' >
               <label>{'Personal Income'}</label>
-              <CurrencyInput className='col-md-12' id='income'
+
+              { <Cleave placeholder='$0'
```
ratneshraval: We should create new currency input component so we can reuse these settings letter.
Let's plan on it when we have another field needing currency.

lauherkcambria: sounds good -- 😄  only reason why i didn't abstract it out yet is because we're only using it one place instead of 2. for dates PR today i will for sure abstract it. 👍

app/javascript/rfa_forms/rfa01a_edit_view/employmentCard.jsx
```diff
@@ -50,14 +52,22 @@ export default class Employment extends React.Component {
...
-                prefix='$'
-                precision='0'
-                maxLength={11}
-                onChangeEvent={(event) => this.onEmploymentChange('income', event.target.value)} />
+                options={{
+                  prefix: '$',
+                  numeral: true,
+                  numeralThousandsGroupStyle: 'thousand',
+                  numeralDecimalScale: 0,
+                  numeralIntegerScale: 10,
```
lavanyareddyreddy: character length should be 8 but it is taking 10. can you double check.

lauherkcambria: good catch

PR #183, opened 2017-09-25
Validator v3

ratneshraval:
- https://www.pivotaltracker.com/story/show/151420115
- https://www.pivotaltracker.com/story/show/151302409

- validator update. Errors added to state
- phone number field allow to backspace when cursor is after area code.


anuroop15:
screen shot 2017-09-26 at 9 28 00 am

@ratneshraval : was this "Unknown" wrapper tag over input created by Cleave.js ?? Can this be named according to type of field it's wrapping.?

anuroop15:
screen shot 2017-09-26 at 9 31 56 am

good work on the bug to delete numbers before parentheses ().  I have identified that  hitting backspace once removes 2 numbers at a time when removing (123)- ___-___ 3 from the field.

ratneshraval:
@anuroop15 
1. unknown tag is only under react code tab. If you look actual HTML, there is no unknown tag. 
2. I am aware of this bug and that is a problem with cleave (https://github.com/nosir/cleave.js/issues/234). I think this bug is better than not allowing to delete at all. We will have a bug story for that where we can research more and fix.

app/javascript/components/common/phoneNumberFields.js
```diff
@@ -5,44 +5,44 @@ import {DropDownField} from './dropDownField'
...
 
-    const phoneNumberId = this.props.idPrefix + 'number'
-
     return (
       <div>
 
         <CleaveInputField
           gridClassName='col-md-4'
-          id={phoneNumberId}
+          id={this.phoneNumberId}
```
lauherkcambria: why make this change?

ratneshraval: so I don't have to build phoneNumberId twice. Before we were building the string in constructor and render. Now I build it in constructor and use it in render.

PR #187, opened 2017-10-03
Adding docker compose

ratneshraval:
- https://www.pivotaltracker.com/story/show/151624427

- Setup to run rspec inside container.

- Affects Jenkins PR testing


lib/tasks/spec.rake
```diff
@@ -0,0 +1,52 @@
...
+end
+
+namespace :spec do # rubocop:disable BlockLength
+  def file_list
+    # first ARGV is task name
+    args = ARGV.drop(1)
+    args.any? ? args.join(' ') : 'spec'
+  end
+
+  def webpack?
```
lavanyareddyreddy: I think this can be renamed because its not returning a boolean

ratneshraval: Updated.

docker-compose.yml
```diff
@@ -0,0 +1,36 @@
...
+# 6/23/2017 - version 3.x does not support extends yet.
+# - https://docs.docker.com/compose/extends/#extending-services
+services:
+  cals:
+    build: .
+    command: bash -c "bundle check || bundle install && yarn && bundle exec puma"
+    ports:
+      - "3000:3000"
+      - "4857:4857"
+    environment:
```
anuroop15: @ratneshraval : if docker file shows all of our env files, doesn't our env files get exposed ?

ratneshraval: It is ok for test setup.

PR #188, opened 2017-10-03
Implementing global header

anuroop15:
- yes

- https://www.pivotaltracker.com/story/show/150495627
-

- NA

-

-

-


app/javascript/header/index.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import {GlobalHeader} from 'react-wood-duck'
+
+export default class HeaderComponent extends React.Component {
+  render () {
+    return (
+      <div>
```
ratneshraval: Why extra div?

anuroop15: Added div to wrap globalHeader component and future components if any

PR #174, opened 2017-09-14
 Child desired card

anuroop15:
#### What tests were added (if relevant), and do all tests (rspec, karma, and integration CALS tests) pass?
- yes

#### Which User story does this relate to?
- yes

#### Brief feature description
-https://www.pivotaltracker.com/story/show/147207475
https://www.pivotaltracker.com/story/show/147955307
https://www.pivotaltracker.com/story/show/147278555

#### Does this PR block/effect other stories?
- no

#### Do any commands needs to be run in order to run the project with this PR?
- npm run karma, ./bin/server

#### Code coverage percent diff as a result of this PR?
- 76.6%

#### JS errors diff as a result of this PR?
-  76

PR #164, opened 2017-08-29
Last name validation

anuroop15:
#### What tests were added (if relevant), and do all tests (rspec, karma, and integration CALS tests) pass?
- yes

#### Which User story does this relate to?
-#150017724, #150017723, #150017727

#### Brief feature description
- covers validations first and last name before saving the application.

#### Does this PR block/effect other stories?
-N/A

#### Do any commands needs to be run in order to run the project with this PR?
- npm run dev

#### Code coverage percent diff as a result of this PR?
- 75.8%

#### JS errors diff as a result of this PR?
- 71

PR #191, opened 2017-10-06
updated onSelection function

anuroop15:
- yes

- https://www.pivotaltracker.com/story/show/148393357

-  Auto fill zip, city and state dropdown with smartyCity API

- NA

- npm run dev

- 76%

- 65


app/javascript/rfa_forms/rfa01a_edit_view/applicantMaritalHistoryCardGroup.js
```diff
@@ -171,7 +171,6 @@ export default class ApplicantMaritalHistoryCardGroup extends React.Component {
                         nameTypes={this.props.nameTypes}
                         stateTypes={this.props.stateTypes}
                         marriageTerminationReasons={this.props.marriageTerminationReasons}
-                        validator={this.props.validator}
```
ratneshraval: Why is this validator removed?

anuroop15: It's duplicated twice in line #164 and line #174. So removed one

PR #200, opened 2017-10-17
Dictionaries for facilities

Nasir2712:
-

--https://www.pivotaltracker.com/n/projects/1662141/stories/151925127
--https://www.pivotaltracker.com/n/projects/1662141/stories/147304283

-pull county and facility type dictionaries and show results

-

-

-

-


app/javascript/search/search.js
```diff
@@ -59,21 +59,21 @@ export default class Search extends React.Component {
     this.state.inputData = DataSearch
 
     // call http request function with arguments
-    fetchRequest('/facilities/search', 'POST', params).then(
+    var url = urlPrefixHelper('/facilities/search')
```
ratneshraval: FetchRequest method already appends urlPrefix. You don't need it here.

Nasir2712: addressed

app/javascript/search/search.js
```diff
@@ -59,7 +59,7 @@ export default class Search extends React.Component {
     this.state.inputData = DataSearch
 
     // call http request function with arguments
-    var url = urlPrefixHelper('/facilities/search')
+    var url = '/facilities/search'
```
ratneshraval: Should use let instead of var.

PR #212, opened 2017-10-23
fix unknown attr issue

lavanyareddyreddy:
-no

-https://www.pivotaltracker.com/story/show/152120208

-ignore attributes not defined in my Rails models

-no

-no

-n/a

-n/a


app/models/cals_base.rb
```diff
@@ -15,5 +15,7 @@ def self.parent_path
     ''
   end
 
-
+  def self.new(*attributes)
+    super(attributes.first.reject{|k,v| !self.instance_methods.include?(k.to_sym) }) unless attributes.first.class == String
```
ratneshraval: Why attributes.first ?

lavanyareddyreddy: Because we are using a splat operator and that collects all params that are passed to that method and returns all params in an array. Although we just pass hash there are some other cases where multiple Params are being passed e.g login page params.

ratneshraval:
Jenkins retest this please.

ratneshraval:
You can get rid of if/else block and keep it direct
```
def self.new(*attributes)
    super(attributes.first.reject{|k,v| !self.instance_methods.include?(k.to_sym) })
end
```

For that you need to adjust User class and AuthenticationProvider.

ratneshraval:
User.rb 

```
class User < CalsBase
  attr_accessor :user, :staffId, :roles, :county_code, :county_name, :privilege

  # def initialize(user, staffId, roles, county_code, county_name, privilege)
  #    @user = user
  #    @staffId = staffId
  #    @roles = roles
  #    @county_code = county_code
  #    @county_name = county_name
  #    @privilege = privilege
  # end
end

```

AuthenticationProvider

```
 def get_user_from_session
    User.new(JSON.parse(session[:user_details]))
    # user_details = JSON.parse(session[:user_details])
    # User.new(user_details['user'],
    #          user_details['staffId'],
    #          user_details['roles'],
    #          user_details['county_code'],
    #          user_details['county_name'],
    #          user_details['privilege'])
  end
```

lavanyareddyreddy:
Jenkins retest this please.

PR #227, opened 2017-11-02
resolve lint errors

ratneshraval:
Remove all lint errors


app/javascript/helpers/cardsHelper.jsx
```diff
@@ -19,9 +19,9 @@ export const removeCardWithId = (inputArray, index, newCardFields) => {
   let inputList = Immutable.fromJS(inputArray)
   let itemForDeletion = inputList.get(index)
   let visibleCount = inputArray.length
- // the item marked for deletion has already been saved once,
- // so we need to add a flag for deletion to make a delete api call
- // rather than an update api call.
+  // the item marked for deletion has already been saved once,
```
anuroop15: do we need these comments??

ratneshraval: No.

anuroop15: yes, but the code here and comments explaining the code doesn't fit here.

ratneshraval: That's true but it explains why this function is needed instead of the regular removeCard function. We can add more comments to explain the logic.

PR #235, opened 2017-11-09
cleave version update to fix backspace bug

ratneshraval:
- https://www.pivotaltracker.com/story/show/151493186

- cleave js backspace bug with special characters


anuroop15:
is yarn.lock needed?

ratneshraval:
@anuroop15 yes yarn.lock is needed to keep versions in sync on every machine

PR #236, opened 2017-11-09
Improve Code Climate Coverage II

Nasir2712:
-

- Improving cc coverage 

- no

- no

-


app/javascript/rfa_forms/rfa01b_edit_view/index.js
```diff
@@ -9,6 +9,7 @@ import CrimeBackgroundAgainstCohabitant from './crimeBackgroundAgainstCohabitant
 import PrivacyStatement from './privacyStatement'
```
Nasir2712: This file's grading is not upgraded because I found out that rfa01b form is still incomplete. Therefore, I stopped making changes to it.

ratneshraval: Sounds good.

PR #243, opened 2017-11-15
Logout from react-wood-duck

Nasir2712:
- https://www.pivotaltracker.com/n/projects/1662141/stories/151454599

-  user can logout

- no

- bundle install, yarn install

-


app/controllers/cals_base_controller.rb
```diff
@@ -3,6 +3,13 @@ class CalsBaseController < ApplicationController
   include RedisHelper
   before_action :authenticate_with_cwds # , unless: 'Rails.env.development?'
 
+
+  def logout
+    delete_user_from_session
+    default_callback_url = ''
```
lauherkcambria: having a default_callback_url here may not be the best way to do this. How about using an env variable like we are in the other methods, i.e. AUTHENTICATION_API_BASE_URL or add a new env variable. these callback urls should be hidden from code base.

Nasir2712: hmm i dont know if you remember this is what i did earlier because I asked you a better way to do this but nesh said it's not needed and then i removed it.

ratneshraval: I remember now.
I agree with what Keith recommends. The reason for us to send blank is so that API can redirect to default login page. Ideally we should never need to send callback_url. So that's why we are just sending blank. 
This may come in handy in future and then we can move it to env var, just my 1 cent :)

lauherkcambria: that's a good cent 😄

PR #244, opened 2017-11-16
Logout bug

Nasir2712:
- https://www.pivotaltracker.com/n/projects/1662141

- user logout bug was not passing props

- no 

- no

-


app/javascript/header/index.js
```diff
@@ -5,7 +5,7 @@ export default class HeaderComponent extends React.Component {
   render () {
     return (
       <div>
-        <GlobalHeader />
+        <GlobalHeader {...this.props}/>
```
ratneshraval: Can you explicitly specify each prop instead of ...props

ratneshraval:
@Nasir2712 can you add a test for headerComponent too?

app/javascript/header/index.js
Nasir2712: yup

Nasir2712:
@ratneshraval  added a test case and addressed your comments.

PR #245, opened 2017-11-17
Rfa01c edit or create on start

Nasir2712:
- https://www.pivotaltracker.com/n/projects/1662141/stories/151572413

- rfa01c edit or create on start rfa01c button

- no

- no

-


app/javascript/rfa_forms/rfa_packet_list/index.js
```diff
@@ -54,6 +55,7 @@ export default class Rfa01PacketList extends React.Component {
             focusComponentName={this.state.focusComponentName}
             applicationId={applicationId}
             childDesired={childDesired}
+            rfa01CForm={rfa01CForm || []}
```
ratneshraval: Why array here instead of object? There is only 1 C form.

app/controllers/rfa/c01_controller.rb
```diff
@@ -1,14 +1,14 @@
 class Rfa::C01Controller < CalsBaseController
 
-  def create
+  def index
     # TODO: GET /rfa-1a-forms/{applicationId}/rfa-1c-forms and create C01 below only if there are no existing applications
```
ratneshraval: TODO is done now, right?

app/javascript/rfa_forms/rfa_packet_list/rfa01cOverview.js
```diff
@@ -11,8 +11,9 @@ const blankChildIdentified = Object.freeze({
 export default class Rfa01COverview extends React.Component {
   render () {
     let applicationId = this.props.applicationId
-    let childDesired = this.props.childDesired
-    let childIdentified = childDesired.child_identified
+    let rfa01CForm = this.props.rfa01CForm
+    let rfc01Link = rfa01CForm.length > 0 ? '/rfa/a01/' + this.props.applicationId + '/c01/' + this.props.rfa01CForm[0].id + '/edit' : '/rfa/a01/' + this.props.applicationId + '/c01/'
```
ratneshraval: Please add test case to catch these condition. It is really close to 100% coverage for the file

app/javascript/rfa_forms/rfa_packet_list/index.js
lauherkcambria: what do you think about using default props here instead of `{rfa01CForm || []}` (will need to change array to object or whatever when that api fix gets in)

app/javascript/rfa_forms/rfa_packet_list/rfa01cOverview.js
lauherkcambria: nice use of .length for presence check. 😄  Instead of using `this.props.rfa01CForm[0].id` and `this.props.applicationId` lets just use the `applicationId` and `rfa01CForm` variables defined right above this line.

app/javascript/rfa_forms/rfa_packet_list/rfa01cOverview.js
```diff
@@ -21,10 +22,9 @@ export default class Rfa01COverview extends React.Component {
         handleOnClick={() => this.props.setFocusState('Rfa01COverview')}
         focusClassName={this.props.getFocusClassName('Rfa01COverview') + ' ' + 'card phone-section double-gap-top'}>
         {childIdentified &&
-          <form action={urlPrefixHelper('/rfa/a01/' + this.props.applicationId + '/c01/')} method="post">
-            <input type="hidden" name="authenticity_token" value={getCsrfToken('csrf-token')} />
-            <button className="btn btn-default" type='submit'><p>Start RFA 01 C</p></button>
-          </form>
+          <a href={urlPrefixHelper(rfc01Link)} className='btn btn-default'>
+            <p>Start RFA 01 C </p>
+          </a>
```
lauherkcambria: nice -- just a thought here, what do you think about abstracting this into a link component?

app/controllers/rfa/c01_controller.rb
```diff
@@ -1,14 +1,14 @@
...
     rfa_c01_app_response = rfa_c01_application_helper.create(params[:a01_id], '{}')
     redirect_to edit_rfa_a01_c01_path(id: rfa_c01_app_response.id, a01_id: params[:a01_id])
   end
 
-  def index
-    @dictionaries = dictionaries_helper.rfa_c01_dictioniaries
-  end
+  #def index
+    #@dictionaries = dictionaries_helper.rfa_c01_dictioniaries
+  #end
```
lauherkcambria: do we need to keep this as a comment?

app/javascript/rfa_forms/rfa_packet_list/index.js
lauherkcambria: we're in index here keith, duh

Nasir2712: api needs to fix it

ratneshraval: Even though it is index, we can still add propTypes and defaultProps to address what Keith suggested.

app/javascript/rfa_forms/rfa_packet_list/rfa01cOverview.js
```diff
@@ -11,21 +10,17 @@ const blankChildIdentified = Object.freeze({
 export default class Rfa01COverview extends React.Component {
   render () {
     let applicationId = this.props.applicationId
-    let childDesired = this.props.childDesired
-    let childIdentified = childDesired.child_identified
+    let rfa01CForm = this.props.rfa01CForm
+    let apiUrlPath = rfa01CForm.length > 0 ? '/rfa/a01/' + applicationId + '/c01/' + rfa01CForm[0].id + '/edit' : '/rfa/a01/' + applicationId + '/c01/'
```
ratneshraval: This comment is not requesting change. An observation,
Since you have dedicated Rfa01cLink component, it almost makes more sense to move this conditional inside that. What's the thought @Nasir2712 @lauherkcambria

app/javascript/rfa_forms/rfa_packet_list/index.js
Nasir2712: I've tested defaultProps to only work when we have undefined values as props. Do you know if defaultProps will work for null values? In this specific case the rfa01CForm is null so I have to test and see if defaultProps will work or not.

ratneshraval: You're right. That won't work. Keep it as is.

app/javascript/rfa_forms/rfa_packet_list/rfa01cOverview.js
Nasir2712: i think i would like to move that to another component just because it would be much easier to test that. No?

ratneshraval: Go for it~

Nasir2712: Cool

app/javascript/rfa_forms/rfa_packet_list/index.js
anuroop15: we can create an extra layer to escape empty array usage

anuroop15:
Missing test cases for /rfa01cLink.js file.

Nasir2712:
@ratneshraval completed the test cases with 100% on branches

app/javascript/rfa_forms/rfa_packet_list/rfa01cCreateLink.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import {urlPrefixHelper} from 'helpers/url_prefix_helper.js.erb'
+
+const Rfa01cCreateLink = ({
+  applicationId,
+  rfa01CForm
+}) => {
+  return (
+    <div id={'rfa01cCreate'}>
```
ratneshraval: if you have string value, you don't need to put { } around it. `id={'rfa01cCreate'}` should be `id='rfa01cCreate'`

app/javascript/rfa_forms/rfa_packet_list/rfa01cCreateLink.js
```diff
@@ -0,0 +1,23 @@
...
+      <a href={urlPrefixHelper('/rfa/a01/' + applicationId + '/c01/')} className='btn btn-default'>
```
ratneshraval: It makes sense to have a link component so we don't have to worry about urlPrefixHelper import.

PR #249, opened 2017-11-21
Polyfills import for ie11 CALS-4268 CALS-4266

ratneshraval:
- CALS-4266 https://osi-cwds.atlassian.net/browse/CALS-4266
- CALS-4268 https://osi-cwds.atlassian.net/browse/CALS-4268

- polyfill for IE 11 

- No

- No


anuroop15:
@ratneshraval : import 'babel-polyfill' individually is better that importing form packs. As babel-polyfill was available globally, we don't need extra polyfill.js file in packs.

ratneshraval:
@anuroop15 the idea is to have all polyfills with single import. If we have to import another item, now we can just import it in 1 file and it will automatically get imported everywhere.

anuroop15:
i think webpack config file should clear our problem with ie promise and tests. I don't have a solution now but the present import is not a proper practice.

ratneshraval:
Sure, I agree we should see if there is a different solution to this.
The polifills.js file is already part of the webpack entry and it is still not resolving the issue. But this solution works.

PR #251, opened 2017-11-21
fix HTTP request error while testing feature spec

lavanyareddyreddy:
- https://osi-cwds.atlassian.net/browse/CALS-4853

- fix HTTP request error while testing feature spec

- no

- rspec spec or rspec spec/features

- n/a


lavanyareddyreddy:
Jenkins, retest this please.

spec/support/vcr_setup.rb
```diff
@@ -16,6 +16,6 @@
   # end
 
   # :host, :headers
-  c.default_cassette_options = {match_requests_on: [:path, :method, :body]}
+  c.default_cassette_options = {match_requests_on: [:path, :method]}
```
ratneshraval: Why aren't we matching on body?

lavanyareddyreddy: because we use faker and body changes with every test.

ratneshraval: Which tests are affected? Since we are using cassettes, shouldn't the body always be same?

lavanyareddyreddy: Any tests that does post with faker data is affected.  The response body will always be the same for a matched request.

ratneshraval: My concern is, if we have 2 requests with same path, method but different body, it may not locate correct cassette.

ratneshraval:
Coverage shows 87% now with this change.

spec/support/vcr_setup.rb
lavanyareddyreddy: I would say thats an exception because each scenario should be testing specific use case(and thats probably the reason why default match on `match_requests_on` is [method and uri](https://relishapp.com/vcr/vcr/v/1-5-0/docs/configuration/default-cassette-options). 

And if/when we have such case we can then use something like this on that test `use_vcr_cassette('mycassette', :match_requests_on => [:method, :host, :path])` and avoid using faker data on that test so that we don't see these warnings on it.

lavanyareddyreddy:
updated code coverage from 83 to 87.

PR #253, opened 2017-11-22
add address geoservice to all card

anuroop15:
Add Address Component in RFA_01A and RFA_01C


app/javascript/components/rfa_forms/addressComponent.js
```diff
@@ -41,17 +43,19 @@ export default class AddressComponent extends React.Component {
...
-        addressType={this.props.addressType}
-        id={this.props.id}
-        suggestions={this.state.suggestions}
-        addressFields={this.props.addressFields}
-        stateTypes={this.props.stateTypes}
-        onChange={this.props.onChange}
-        onSuggestionsFetchRequested={this.onSuggestionsFetchRequested}
-        onSuggestionSelected={this.onSuggestionSelected}
-      />
+      <ErrorBlock>
```
ratneshraval: I like the idea of error block but I don't know if this is the right place to use it.

app/javascript/components/rfa_forms/addressComponent.js
```diff
@@ -32,7 +33,8 @@ export default class AddressComponent extends React.Component {
     let updateSuggetions
     fetchRequest(url, 'POST', params).then(
       response => response.json()).then((response) => {
-      updateSuggetions = response[0]
+      updateSuggetions = response[0].street_address ? response[0] : suggestion
```
ratneshraval: Is this workaround till the bug is fixed by TPT3?

app/javascript/rfa_forms/rfa01a_edit_view/applicantMaritalHistoryCardGroup.js
```diff
@@ -120,6 +121,15 @@ export default class ApplicantMaritalHistoryCardGroup extends React.Component {
     this.changeAdultChild('address', address, index)
   }
 
+  onSelection (autofillData, index) {
```
ratneshraval: Now that we have this onSelection function in 3 places, probably good idea to refactor it

app/javascript/components/rfa_forms/addressComponent.js
anuroop15: I think this is the place to start, as we are relying on geoservice api which was different service. and any breakage in this service shouldn't break out entire page.

app/javascript/components/rfa_forms/addressComponent.js
anuroop15: yes. once they fix the issue, we doesn't need this extra conditional render

app/javascript/rfa_forms/rfa01a_edit_view/applicantMaritalHistoryCardGroup.js
anuroop15: sure. do you want me to do that in this PR?

ratneshraval:
`rfa01c_edit_view/desiredChildcard.jsx` is rated as F :(

ratneshraval:
jenkins, retest this please.

app/javascript/rfa_forms/rfa01a_edit_view/applicantMaritalHistoryCardGroup.js
ratneshraval: @Nasir2712 this function will need to be updated, right?

Nasir2712: ya i did

app/javascript/rfa_forms/rfa01a_edit_view/employmentCard.jsx
```diff
@@ -35,7 +35,11 @@ export default class Employment extends React.Component {
     newData = newData.setIn(['physical_address', key], value)
     this.props.setParentState('employment', newData.toJS())
   }
-
+  onSelection (autofillData, employmentProps) {
```
lauherkcambria: is there no way to have this methods in the addressComponent itself?

app/classes/faraday/faraday_base.rb
```diff
@@ -23,16 +23,19 @@ def self.delete(url, auth_header)
...
+      Rails.logger.info('API call request: ')
+      Rails.logger.info("URL : #{url}")
+      Rails.logger.info("Method : #{method}")
+    end
 
     response = Faraday.send(method) do |req|
       req.url url
       req.headers = default_headers(auth_header)
       req.body = body if method.in?(BODY_METHODS)
+      req.options.timeout = 5
```
lauherkcambria: why 5? is this a project wide mandated standard? let's check.

app/javascript/components/common/errors/errorBlock.jsx
```diff
@@ -0,0 +1,34 @@
...
+            {this.state.error && this.state.error.toString()}
+            <br />
+            {this.state.errorInfo.componentStack}
+          </details>
+        </div>
+      )
+    }
+    return this.props.children
+  }
+}
```
lauherkcambria: nice! this will always render wherever the component that is throwing an error is, however.  Is this spec coming from design? I feel that usually when there is an error, a banner is thrown, are there design specs for this?

app/javascript/components/common/errors/errorBlock.jsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+
+export default class ErrorBlock extends React.Component {
+  constructor (props) {
+    super(props)
+    this.state = { error: null, errorInfo: null }
```
lauherkcambria: how about using default props and proptypes here

app/javascript/rfa_forms/rfa01a_edit_view/employmentCard.jsx
Nasir2712: let me give this a try.

app/javascript/components/common/errors/errorBlock.jsx
Nasir2712: i am not sure if we are gonna use this or not. It's just that we still didn't figure out a better way to implement this. So that's why the component exists but I am not rendering it any of the files for now.

app/classes/faraday/faraday_base.rb
Nasir2712: so nesh and i were just testing the timeout. There is no standard set for now. If you have any set standard let me know. I'll update it

lauherkcambria: sounds good, we should probably follow up on this.

ratneshraval: I will follow up on this, 5 may be too short but we do not have any lengthy calls, so it should work in the mean time.

app/javascript/components/rfa_forms/addressComponent.js
```diff
@@ -2,6 +2,8 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import {CommonAddressComponent} from 'react-wood-duck'
 import {fetchRequest} from 'helpers/http'
+import ErrorBlock from 'components/common/errors/errorBlock.jsx'
```
ratneshraval: should remove this import.

Nasir2712:
@ratneshraval and @keith let me know if you guys want me to make changes to variable names in onSelection method in address component and props too.

app/javascript/rfa_forms/rfa01a_edit_view/employmentCard.jsx
```diff
@@ -25,19 +26,18 @@ const blankEmploymentFields = Object.freeze({
...
 export default class Employment extends React.Component {
   onEmploymentChange (key, value) {
-    let newData = Immutable.fromJS(this.props.employment || blankEmploymentFields)
+    let newData = Immutable.fromJS(this.props.employment)
     newData = newData.set(key, value)
     this.props.setParentState('employment', newData.toJS())
   }
   onPhysicalAddressChange (key, value) {
-    let newData = Immutable.fromJS(this.props.employment || blankEmploymentFields)
+    let newData = Immutable.fromJS(this.props.employment)
```
lauherkcambria: nice! 👍

app/javascript/components/rfa_forms/addressComponent.js
```diff
@@ -26,17 +28,27 @@ export default class AddressComponent extends React.Component {
       })
     })
   }
+
+  onSelectionUpdateProp (updateSuggetions) {
```
lauherkcambria: nice! 👍

Nasir2712:
@ratneshraval I refactored and added address component to all the files wherever we are using address fields that includes a,b and c forms. I made sure they are working and are being loaded once they are saved. 

Final Update - I am using two onSelection methods one in addressCard and one in employment card.

ratneshraval:
@Nasir2712 all addresses work and save correctly when there is a valid address. But when you input an invalid address, you get a console error and then page crashes on next attempt to input.

Nasir2712:
@ratneshraval So I was able to figure out uncontrolled input warning and I fixed that warning. But this error is tricky to figure out - Objects are not valid as a React child (found: object with keys {value, id}). If you meant to render a collection of children, use an array instead. 

I feel like React is trying to evaluate the state object as a whole and not as object.id or object.value, that's why the error but I was not able to figure out why this doesn't happen when we input the right address. It's kinda confusing. Let's discuss in detail tomorrow.

PR #254, opened 2017-11-27
Languages multiselect in residence card

Nasir2712:
- https://osi-cwds.atlassian.net/browse/CALS-4393

-  create user interface for "Languages" section in residence card (RFA 01A)

- no

- yarn install

-


app/javascript/rfa_forms/rfa01a_edit_view/aboutThisResidenceCard.jsx
```diff
@@ -90,15 +91,26 @@ export default class AboutThisResidenceCard extends React.Component {
...
 
-            <DropDownField id='languagesSpoken' gridClassName='col-md-12'
+            <MultiSelect
+              label="Language(s) spoken in the home"
+              value={aboutResidence.home_languages.map((hl) => hl.value)}
+              className="languages"
+              optionList={this.props.languageTypes.map((type) => ({label: type.value, value: type.value, id: type.id}))}
+              onChange={(event) => this.props.setParentState('home_languages', event.map((e) => ({id: e.id, value: e.value})))}/>
+
+            {/* <DropDownField id='languagesSpoken' gridClassName='col-md-12'
```
ratneshraval: Remove this commented dropdown

app/javascript/rfa_forms/rfa01a_edit_view/aboutThisResidenceCard.jsx
```diff
@@ -19,7 +20,7 @@ export default class AboutThisResidenceCard extends React.Component {
   }
 
   onChange (key, value) {
-    let othersMailing = Immutable.fromJS(this.props.aboutResidence.other_people_using_residence_as_mailing || [othersUsingAddressMailing])
+    let othersMailing = Immutable.fromJS([this.props.aboutResidence.other_people_using_residence_as_mailing])
```
ratneshraval: If you have `aboutResident.body_of_water_exists` then defaultProps won't kick in for `.other_people_using_residence_as_mailing`, and it will still be null.

app/javascript/rfa_forms/rfa01a_edit_view/aboutThisResidenceCard.jsx
```diff
@@ -90,15 +91,26 @@ export default class AboutThisResidenceCard extends React.Component {
...
-              onChange={(event) => this.props.setParentState('home_languages', [dictionaryNilSelect(event.target.selectedOptions[0])])} />
+              onChange={(event) => this.props.setParentState('home_languages', [dictionaryNilSelect(event.target.selectedOptions[0])])} /> */}
           </form>
         </div>
       </div>
     )
   }
 }
+
+AboutThisResidenceCard.defaultProps = {
```
ratneshraval: Can you add other propTypes while you are here :)

app/javascript/rfa_forms/rfa01a_edit_view/residenceCardsMain.jsx
```diff
@@ -11,7 +11,10 @@ const blankResidenceFields = Object.freeze({
   body_of_water_description: '',
   others_using_residence_as_mailing: '',
   directions_to_home: '',
-  home_languages: []
+  home_languages: [{
+    id: '',
```
ratneshraval: Will this fail if you don't pick a home language at all?

app/javascript/rfa_forms/rfa01a_edit_view/residenceCardsMain.jsx
```diff
@@ -64,3 +67,7 @@ export default class ResidenceCards extends React.Component {
     )
   }
 }
+
+ResidenceCards.defaultProps = {
+  residence: blankResidenceFields
```
ratneshraval: 👍

app/javascript/rfa_forms/rfa01a_edit_view/residenceCardsMain.jsx
Nasir2712: Yup because I am removing a conditional on the mutliselect value. I am just using the value from the props. Since I am using undefined i want the default props to work on my multiselect so that it won't fail.

app/javascript/rfa_forms/rfa01a_edit_view/aboutThisResidenceCard.jsx
Nasir2712: sure

app/javascript/components/common/multiSelect.js
```diff
@@ -0,0 +1,36 @@
...
+  className,
+  onChange
+}) => {
+  return (
+    <div className='col-md-12'>
+      <label>{label}</label>
+      <Select
+        className={className}
+        tabSelectsValue={false}
+        multi
```
anuroop15: can we make this component switch from multi select and single select. as we are implementing as common reuasble component.

app/javascript/components/common/multiSelect.js
```diff
@@ -0,0 +1,36 @@
...
+    </div>
+  )
+}
+
+MultiSelect.propTypes = {
+  label: PropTypes.string,
+  value: PropTypes.arrayOf(PropTypes.string),
+  onChange: PropTypes.func
+}
+
```
anuroop15: can we add default Props for react-select

app/javascript/components/common/multiSelect.js
```diff
@@ -0,0 +1,36 @@
...
+      <Select
+        className={className}
+        tabSelectsValue={false}
+        multi
+        searchable={false}
+        value={value}
+        options={optionList}
+        onChange={onChange}
+        clearable={false}
+        placeholder=''
```
anuroop15: adding props is better than leaving empty string value

app/javascript/rfa_forms/rfa01a_edit_view/aboutThisResidenceCard.jsx
```diff
@@ -90,15 +91,26 @@ export default class AboutThisResidenceCard extends React.Component {
               label='Please provide directions, including major cross-street information, to your physical adress.' placeholder=''
               onChange={(event) => this.props.setParentState('directions_to_home', event.target.value)} />
 
-            <DropDownField id='languagesSpoken' gridClassName='col-md-12'
+            <MultiSelect
```
anuroop15: adding id to react elements will differentiate elements If there are multiple multiSelect tags in same page/rfa_01A

app/javascript/components/common/multiSelect.js
Nasir2712: @ratneshraval I am using multi prop in the select component but not passing it as props. If we change it's value to false then you can use it as normal drop down select. What do you think we will use this component for in future mostly? Are we open to using this component as single select dropdown?

ratneshraval: @anuroop15 @Nasir2712 I do not think we need to make that change here. If needed, we will work on making this update.

PR #255, opened 2017-11-29
Feature specs for rfa01c form

lavanyareddyreddy:
-

- feature spec for RFA01 C

- no
 
- rspec spec

- 88.85%


spec/features/rfa_01c_application_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    click_button 'Create RFA Application (Form 01)'
+    expect(page).to have_content 'Rfa-01A Section Summary'
+    page.find('#Rfa01AOverview').find('a.btn.btn-default').click
+    expect(page).to have_content 'Applicant 1 - Information'
+    fill_in('firstname', with: Faker::Name.name, :match => :prefer_exact)
+    fill_in('middleName', with: 'k', :match => :prefer_exact)
+    fill_in('lastName', with: Faker::Name.name, :match => :prefer_exact)
+    find(:select, 'child_financially_supported').first(:option, 'Yes').select_option
+    click_button('Save Progress')
+    byebug
```
Nasir2712: can you remove byebug from here?

lavanyareddyreddy: I dont see any byebug

spec/features/rfa_01c_application_spec.rb
```diff
@@ -6,30 +6,30 @@
...
+    packet_url = url.path.gsub('/edit', '') + '/packet'
     fill_in('firstname', with: Faker::Name.name, :match => :prefer_exact)
     fill_in('middleName', with: 'k', :match => :prefer_exact)
     fill_in('lastName', with: Faker::Name.name, :match => :prefer_exact)
-    find(:select, 'child_financially_supported').first(:option, 'Yes').select_option
+    find(:select, 'child_identified').first(:option, 'Yes').select_option
+    find(:select, 'child_in_home').first(:option, 'Yes').select_option
     click_button('Save Progress')
-    byebug
+    visit page.driver.current_url
```
Nasir2712: I think line 13 & 14 should be moved here

spec/features/rfa_01c_application_spec.rb
Nasir2712: ya it was before you committed. i came in saw reviews needed thought you were done.

spec/features/rfa_01c_application_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    visit packet_url
+    page.find('#Rfa01COverview').find('a.btn.btn-default').click
+    fill_in('firstname', with: Faker::Name.name, :match => :prefer_exact)
+    fill_in('middleName', with: 'k', :match => :prefer_exact)
+    fill_in('lastName', with: Faker::Name.name, :match => :prefer_exact)
+    fill_in('street_address', with: 'address here', :match => :prefer_exact)
+    find(:select, 'name_suffix').first(:option, 'Sr').select_option
+    find(:select, 'grade').first(:option, 'TK').select_option
+    find(:select, 'state_type').first(:option, 'Alaska').select_option
+    fill_in('Zip', with: 'Alaska', :match => :prefer_exact)
```
Nasir2712: @lavanyareddyreddy @ratneshraval  this is actually interesting. So type in the input field for zip accepts text that's why this is passing after I add an expect statement. Not you're fix but just putting it out there.

ratneshraval: I think it is good idea to change this to numbers

spec/features/rfa_01c_application_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    find(:select, 'gender').first(:option, 'Male').select_option
+    fill_in('name_of_school', with: 'School name', :match => :prefer_exact)
+    click_button('Save Progress')
+    visit page.driver.current_url
+    expect(find_field('gender').value).to eq '1'
+    expect(find_field('name_of_school').value).to eq 'School name'
```
ratneshraval: after page refresh, we should verify each item that was inputted for save.

lavanyareddyreddy: good point, verifying each field after save now.

ratneshraval: Check how we can verify fields filled in my faker.

lavanyareddyreddy: Sure,

ratneshraval:
I am approving this since the tests will verify functionality of rails and react side. But we need to think about how to utilize faker between VCR based vs end-to-end tests.

ratneshraval:
jenkins, retest this please.

PR #256, opened 2017-12-01
fix ie dropdown bug with test case

Nasir2712:
- https://osi-cwds.atlassian.net/browse/CALS-4272

- dropdown does not retain selection on ie

- no

-no

-


test/javascript/rfa_forms/rfa01a_edit_view/aboutThisResidenceCard_tests.js
```diff
@@ -63,7 +63,7 @@ describe('Verify Physical Address', function () {
 
   it('verify residence types address change', () => {
     let relationShipField = residenceCardComp.find('#residenceTypes')
-    relationShipField.simulate('change', {target: {selectedOptions: [{value: '1', text: 'Own'}]}})
+    relationShipField.simulate('change', {target: {options: {'1': {value: '1', text: 'Own'}, selectedIndex: 1}}})
```
ratneshraval: When you refactor this, probably good idea to declare this object  `{options: {'1': {value: '1', text: 'Own'}, selectedIndex: 1}` somewhere and reuse it with imports so you don't have to keep repeating it.
Also, if you do that, have more than 1 value in it.

PR #269, opened 2017-12-11
Multiselect Test Cases

Nasir2712:
- N/A

- passing test cases for multiselect

- n/a

- no

-


app/javascript/components/common/multiSelect.js
```diff
@@ -32,7 +32,7 @@ MultiSelect.propTypes = {
   label: PropTypes.string,
   values: PropTypes.array,
   optionList: PropTypes.array,
-  classNameId: PropTypes.string,
+  className: PropTypes.string,
```
ratneshraval: Im surprised it doesn't throw error since className is a key name in itself.

PR #273, opened 2017-12-13
Updating button tag to button component.

nasar-cambria:
- No Story

- Removing button tag and importing button component.

- No

- No

- No


app/javascript/rfa_forms/rfa01a_edit_view/FosterCareHistoryFields.js
```diff
@@ -236,6 +240,7 @@ export class FosterCareHistoryFields extends React.Component {
               // Question 2
               }
               <DropDownField gridClassName='col-md-7'
+                id={'q2-select-dropdown'}
```
ratneshraval: You don't need {} around string values

nasar-cambria: removed.

test/javascript/rfa_forms/rfa01a_edit_view/fosterCareHistoryCard_tests.js
```diff
@@ -231,5 +225,60 @@ describe('foster car card tests', function () {
         expect(input.type()).toEqual('input')
       })
     })
+    describe('when Q2 Adoption Field "dropdown is changed', () => {
```
Nasir2712: discussed about test cases with nasar. Waiting for his response.

nasar-cambria: Removing component instance.

test/javascript/rfa_forms/rfa01a_edit_view/fosterCareHistoryCard_tests.js
```diff
@@ -61,7 +61,7 @@ describe('foster car card tests', function () {
       // has 7 yes/no drop down questions
       expect(fosterCareFieldsComponent.find('div[className="row agency-list-item"]').length).toEqual(7)
       // has 6 total add buttons
-      expect(fosterCareFieldsComponent.find('button').length).toEqual(6)
+      expect(fosterCareFieldsComponent.find('button').length).toEqual(0)
```
lauherkcambria: this doesn't sound right

test/javascript/rfa_forms/rfa01a_edit_view/fosterCareHistoryCard_tests.js
```diff
@@ -113,19 +106,16 @@ describe('foster car card tests', function () {
     })
     describe('when "add another facility" is clicked', () => {
       it('calls add facility', () => {
+        let event
```
lauherkcambria: event is undefined?

test/javascript/rfa_forms/rfa01a_edit_view/fosterCareHistoryCard_tests.js
Nasir2712: Since, now the component instance has been removed, you are not testing it so you don't need the it block to say that. 

You can just wrap this 'when Q2 Adoption Field "dropdown is changed' - statement inside an it block and remove the describe block, which would be much easier to read and also understand what exactly it's testing.

app/javascript/rfa_forms/rfa01a_edit_view/FosterCareHistoryFields.js
```diff
@@ -226,7 +227,10 @@ export class FosterCareHistoryFields extends React.Component {
               }
               <div className={agencyQ1List ? '' : 'hidden'}>
                 <div className='text-center'>
-                  <button onClick={(event) => { this.addAgencyCard(event, q1History.agencies, 'foster_care_licenses_q1', 'agencies') }} className='btn btn-default q1'>Add another Agency +</button>
+                  <Button
+                    buttonId='addAgency_q1'
+                    label='Add Another Agency +'
+                    onClick={(event) => { this.addAgencyCard(event, q1History.agencies, 'foster_care_licenses_q1', 'agencies') }} />
```
ratneshraval: There is no need for additional {}. 
`{(event) => this.addAgencyCard(event, q1History.agencies, 'foster_care_licenses_q1', 'agencies')}` should work

test/javascript/rfa_forms/rfa01a_edit_view/fosterCareHistoryCard_tests.js
nasar-cambria: I am using ```describe()``` to describe the context of this test. Since, every drop down is using a different component i.e why ```describe()``` is repeated. If we need to remove the describe I can use a beforeEach for onChangeSpy and make it global for all of the following tests. This change should be updated now.

test/javascript/rfa_forms/rfa01a_edit_view/fosterCareHistoryCard_tests.js
nasar-cambria: If I don't declare it the test fails saying ```event is not defined```

app/javascript/rfa_forms/rfa01a_edit_view/FosterCareHistoryFields.js
nasar-cambria: Done.

test/javascript/rfa_forms/rfa01a_edit_view/fosterCareHistoryCard_tests.js
```diff
@@ -61,7 +61,7 @@ describe('foster car card tests', function () {
       // has 7 yes/no drop down questions
       expect(fosterCareFieldsComponent.find('div[className="row agency-list-item"]').length).toEqual(7)
       // has 6 total add buttons
-      expect(fosterCareFieldsComponent.find('button').length).toEqual(6)
+      expect(fosterCareFieldsComponent.find('div[className="hidden"] div[className=""] div[className="text-center"]').length).toEqual(6)
```
ratneshraval: So why this `div[className="hidden"]` ? I haven't tried on my computer yet but seems odd

lauherkcambria: because we hide the 'add another agency' button when there isn't anything selected, and this tests that we have 6 hidden buttons on this page, which is correct.

ratneshraval: That makes sense.

PR #274, opened 2017-12-14
Fixed bugs for facility search page

sunilb0575:
-None

-When no value entered/not matched, wasn't showing 'No result banner'.
-There was a error when searched for 'home/house' in facility name field.  

-No

-No

78.39% Statements  62.92% Branches  71.73% Functions  79.32% Lines


app/javascript/search/common/commonUtils.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+
+export const address = (result) => {
+  return result.addresses.length > 0 ? result.addresses[0].address ? result.addresses[0].address.street_address + ',' + result.addresses[0].address.city +
```
ratneshraval: 1. So the use of the function is to build an address, so the name should reflect that, maybe something like `fullAddressString`
2. The function should take a single address and output the string, so it shouldn't be dealing with `result.addresses[0]`, that should be passed in as argument.

app/javascript/search/common/commonUtils.js
```diff
@@ -0,0 +1,10 @@
...
+         ',' + result.addresses[0].address.state + ' ' + result.addresses[0].address.zip_code : 'N/A' : 'N/A'
+}
+
+export const phones = (result) => {
```
ratneshraval: Comments similar to address applies here.

app/javascript/search/search_grid.js
```diff
@@ -40,35 +30,35 @@ export default class SearchGrid extends React.Component {
                 value={result.license_number} />
               <GridInnerLayout
                 title={'Facility Type'}
-                value={result.type.value} />
+                value={result.type ? result.type.value : 'N/A'} />
```
lauherkcambria: maybe we could make a util method for this?

app/javascript/search/common/commonUtils.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+
+export const fullAddressString = (object) => {
+  return object.length > 0 ? object[0].address ? object[0].address.street_address + ',' + object[0].address.city +
```
ratneshraval: An improved approach would be to just pass address object here that will return full string.

app/javascript/search/search_grid.js
```diff
@@ -3,27 +3,17 @@ import PropTypes from 'prop-types'
...
 export default class SearchGrid extends React.Component {
   render () {
     const searchResults = this.props.searchResults
-    const gridResult = searchResults.map((result) => {
-      // let displayAddress = result.addresses[0] !== undefined && result.addresses[0].address.street_address !== undefined
-      let address = result.addresses[0].address.street_address + ',' + result.addresses[0].address.city +
-        ',' + result.addresses[0].address.state + ' ' + result.addresses[0].address.zip_code
-      let phone = result.phones[0].number.replace(/(\d{3})(\d{3})(\d{4})/, '($1)-$2-$3')
+    const gridResult = searchResults.map((result, index) => {
+      const fullAddress = fullAddressString(result.addresses)
```
ratneshraval: And check for `result.addresses.length > 0 ? fullAddressString(result.addresses[0]) : 'N/A'`

ratneshraval: I'm sure you can add another method that handles 'N/A' logic so it isn't a part of your render()

app/javascript/search/search_list.js
```diff
@@ -2,29 +2,28 @@ import React from 'react'
...
+      const phoneNo = createPhoneNumber(result.phones)
 
       return (
-        <tr key={result.license_number}>
+        <tr key={index}>
           <td><a href={urlPrefixHelper('/facilities/' + result.license_number)}>{result.name}</a></td>
           <td>{result.license_number}</td>
-          <td>{result.type.value}</td>
-          <td>{result.status.value}</td>
+          <td>{result.type ? result.type.value : 'N/A'}</td>
```
ratneshraval: Any reason for not using `gridInnerLayoutValue` function here?

app/javascript/search/common/commonUtils.js
```diff
@@ -0,0 +1,14 @@
...
+         ',' + object[0].address.state + ' ' + object[0].address.zip_code : 'N/A' : 'N/A'
+}
+
+export const createPhoneNumber = (object) => {
+  return object.length > 0 ? object[0].number.replace(/(\d{3})(\d{3})(\d{4})/, '($1)-$2-$3') : 'N/A'
+}
+
+export const gridInnerLayoutValue = (object) => {
+  return object ? object.value : 'N/A'
```
ratneshraval: This function should be renamed to reflect that it returns 'N/A' when no value is present.

app/javascript/search/common/commonUtils.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+
+export const addressStringValueOrNa = (addresses) => {
+  if (addresses.length > 0 && addresses[0].address) {
```
Nasir2712: Can ternary operators be used here instead of if else block both here and in phoneNumberOrNa

app/javascript/search/search_grid.js
```diff
@@ -3,27 +3,17 @@ import PropTypes from 'prop-types'
...
 export default class SearchGrid extends React.Component {
   render () {
     const searchResults = this.props.searchResults
-    const gridResult = searchResults.map((result) => {
-      // let displayAddress = result.addresses[0] !== undefined && result.addresses[0].address.street_address !== undefined
-      let address = result.addresses[0].address.street_address + ',' + result.addresses[0].address.city +
-        ',' + result.addresses[0].address.state + ' ' + result.addresses[0].address.zip_code
-      let phone = result.phones[0].number.replace(/(\d{3})(\d{3})(\d{4})/, '($1)-$2-$3')
+    const gridResult = searchResults.map((result, index) => {
+      const fullAddress = addressStringValueOrNa(result.addresses)
```
Nasir2712: I feel like these are unnecessary constants. You can always use addressStringValueOrNa(result.addresses) in the gridInnerLayout component's value to have a common look throughout the file. 

Same in the list file use of constants for the same thing

app/javascript/search/common/commonUtils.js
sunilb0575: Yeah, thought the same. But to make code readable used If-Else.

ratneshraval: I generally think it is more readable to use if-else when there are multiple conditions than a ternary block

app/javascript/search/common/commonUtils.js
```diff
@@ -0,0 +1,29 @@
...
+    return fullAddressString(addresses)
+  } else {
+    return 'N/A'
+  }
+}
+
+export const fullAddressString = (addresses) => {
+  return addresses[0].address.street_address + ',' + addresses[0].address.city + ',' + addresses[0].address.state + ' ' + addresses[0].address.zip_code
```
ratneshraval: There is still a problem here. The function needs to take a single address here, not array of addresses.

app/javascript/search/common/commonUtils.js
```diff
@@ -0,0 +1,29 @@
...
+export const phoneNumberOrNa = (phone) => {
+  if (phone.length > 0) {
+    return createPhoneNumber(phone)
+  } else {
+    return 'N/A'
+  }
+}
+
+export const createPhoneNumber = (phone) => {
+  return phone[0].number.replace(/(\d{3})(\d{3})(\d{4})/, '($1)-$2-$3')
```
ratneshraval: There is still a problem here. The function needs to take a single phone here, not array of phones.

app/javascript/search/search_list.js
```diff
@@ -2,29 +2,28 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import {urlPrefixHelper} from '../helpers/url_prefix_helper.js.erb'
 import {searchDataDefaults} from './search_grid.js'
```
ratneshraval: So you didn't make this change but can you remove this `import {searchDataDefaults} from './search_grid.js'` and make the default prop to match with `search_grid.js`

test/javascript/search/searchList_tests.js
```diff
@@ -83,6 +84,6 @@ describe('Render search list', function () {
   })
   it('Verify assigned Facility ID', function () {
     let tableFaclityID = tableElement.props.children[1].props.children[0]
-    expect(tableFaclityID.key).toEqual('193600008')
+    expect(tableFaclityID.key).toEqual(indexValue)
```
ratneshraval: Rename this `it` block since you're verifying index and not facility id.

app/javascript/search/common/commonUtils.js
Nasir2712: oh got it!

app/javascript/search/search_grid.js
```diff
@@ -3,27 +3,17 @@ import PropTypes from 'prop-types'
...
+      const phoneNo = phoneNumberOrNa(result.phones)
 
       return (
-        <div key={result.license_number} className='grid_view_inner col-xs-12 col-sm-12 col-md-12 col-lg-12' >
+        <div key={index} className='grid_view_inner col-xs-12 col-sm-12 col-md-12 col-lg-12' >
           <div className='col-xs-12 col-sm-1 col-md-1 col-lg-1'>
```
nasar-cambria: So this in not something you changed but why column attribute ```col-xs-12``` is repeated on line 16 and line 17?

PR #277, opened 2017-12-19
format date for save and display bug fix 

lauherkcambria:
- bug fix

-

-

-

-78.94% Statements 
 64.11% Branches
 72.99% Functions 
 79.83% Lines


test/javascript/rfa_forms/rfa01a_edit_view/adultChildrenField_tests.js
```diff
@@ -81,17 +111,29 @@ describe('Adult Children Component', () => {
...
     })
+
+    it('Lives in Home set to false as props', () => {
+      expect(selectLivesInHomeForNotInHomeChild.props().value).toBe('false')
+      expect(adultChildNotInHomeComponent.find('select').length).toEqual(6)
+    })
+
+    it('set lives in Home to true', () => {
+      selectLivesInHome.simulate('change', {target: {value: true}})
+      expect(onChangeSpy).toHaveBeenCalledWith('lives_in_home', true, 0)
```
ratneshraval: Should there be additional expect statement with `length.toEqual(7)` ?

app/javascript/helpers/commonHelper.jsx
```diff
@@ -29,7 +29,10 @@ export const removeLegalNameType = (nameTypes) => {
 export const FormatDateForPersistance = (dateString) => {
   let persistantDateString
   if (dateString) {
-    persistantDateString = SplitDate(dateString, '/', '-')
+    if (dateString.length === 10) {
+      const dateStringArray = dateString.split('/', 3)
+      persistantDateString = ([dateStringArray[2], dateStringArray[0], dateStringArray[1]]).join('-')
```
ratneshraval: I think the moment code didn't work because we didn't have length check in place. That may be something to explore to avoid the calculation ourself.

lauherkcambria: good idea-  we should probably explore that not as a part of this hotfix, however, so i'll make a user story in the backlog to investigate implementing moment.

ratneshraval: Sure, that works for me.

test/javascript/rfa_forms/rfa01a_edit_view/adultChildrenField_tests.js
Nasir2712: I think so too just to be more sure.

lauherkcambria: addressed (note, the test case is 6, not 7 because the previous test was not just selectLIvesInHome)

PR #278, opened 2017-12-19
name type dropdown fix

Nasir2712:
-

- Name Type dropdown in Name Card was not working. This PR is a fix for this bug

-

-

-


ratneshraval:
@Nasir2712 so there should be a new test case catching the bug, right?

Nasir2712:
@ratneshraval so if i remember correctly the test was already failing. I fixed the test case and pushed it i believe in some other PR. There should be a test case for name types in completeNameFields.

ratneshraval:
So in development branch we have this bug and all tests pass, means there is not a specific test that failed for this bug, right?

lauherkcambria:
@Nasir2712 i approved, but lets add the test nesh is speaking of as well 😄 👍

Nasir2712:
@lauherkcambria its already there in development branch

PR #287, opened 2018-01-02
Remove refs from facility search

Nasir2712:
- n/a

- clean refs from facility search
 
- no

- no

- statements: 80,
   branches: 65,
   functions: 74,
   lines: 81


app/javascript/search/stylesheets/search.scss
```diff
@@ -30,6 +30,7 @@ input[type=number]::-webkit-outer-spin-button {
   width: 100%;
   height: 30px;
   padding: 5px;
+  margin-top: -4px;
```
sunilb0575: Good catch!

PR #290, opened 2018-01-03
Update codeclimate to disable similar check.

ratneshraval:
- https://osi-cwds.atlassian.net/browse/CAL1-26

- Code climate file now ignores similar code found
- Identical code errors will still be reported.


.codeclimate.yml
```diff
@@ -1,28 +1,26 @@
-engines:
+version: "2"
+checks:
+  similar-code:
+    enabled: false
```
sunilb0575: So from now on our components will not be checked for similar code?

ratneshraval: Yes

PR #291, opened 2018-01-03
adding children assigned worker first and last name

Nasir2712:
- https://osi-cwds.atlassian.net/browse/CAL1-11

- In facility profile, children section should show assigned worker first name and last name if it has.

- no

-  no

-


app/javascript/search/common/commonUtils.js
```diff
@@ -32,6 +32,14 @@ export const checkForNA = (object) => {
   return (object && object.value) ? object.value : 'N/A'
 }
 
+export const checkfullNameorNA = (object) => {
+  return object ? assignedWorkerFullName(object) : 'N/A'
+}
+
+export const assignedWorkerFullName = (object) => {
```
ratneshraval: if this is `object.first_name` and `object.last_name`, shouldn't it be named something more generic than `assignedWorkerFullName`?

Nasir2712: agreed, any suggestions?

ratneshraval: Sure, let's go with that for now.

PR #293, opened 2018-01-04
Handle on clear conditional bug fix

lauherkcambria:
- https://osi-cwds.atlassian.net/browse/CALS2-5060

- for conditional fields, when the user makes a selection to hide the field, upon reselecting, the data needs to be cleared.  Any data entered into a field that dissapears after a conditional check will be wiped.

- no

- no

-79.69% Statements 
 64.66% Branches
74.54% Functions 
80.64% Lines


app/javascript/rfa_forms/rfa01a_edit_view/FosterCareHistoryFields.js
```diff
@@ -87,6 +87,17 @@ export class FosterCareHistoryFields extends React.Component {
     this.props.setParentState('fosterCareHistory', newerData.toJS())
   }
 
+  handleClearOnConditionalChange (key, subKey, value, hiddenKey, hiddenDefaultValue) {
+    if (value === 'false') {
+      let newData = Immutable.fromJS((Object.keys(this.props.fosterCareHistory).length !== 0) ? this.props.fosterCareHistory : blankFosterCareFields)
```
ratneshraval: The code inside `value === false`looks like it can be extracted out to a common method.

app/javascript/rfa_forms/rfa01b_edit_view/crimeBackgroundAgainstCohabitant.js
```diff
@@ -34,8 +35,8 @@ export default class CrimeBackgroundAgainstCohabitant extends React.Component {
   }
 
   render () {
-    const arrestedForCrime = this.props.arrestedForCrime
-    const disclosures = this.props.disclosures
+    const arrestedForCrime = String(this.props.arrestedForCrime)
```
ratneshraval: What's the reason for wrapping the prop in `String`?

app/javascript/rfa_forms/rfa01b_edit_view/crimeBackgroundAgainstCohabitant.js
```diff
@@ -34,8 +35,8 @@ export default class CrimeBackgroundAgainstCohabitant extends React.Component {
...
+    const disclosures = checkArrayObjectPresence(this.props.disclosures) || [disclosureDefaults]
```
ratneshraval: Is this because `this.props.disclosures` are sometimes a blank array `[]` ?

app/javascript/rfa_forms/rfa01a_edit_view/residenceCardsMain.jsx
```diff
@@ -32,6 +33,19 @@ export default class ResidenceCards extends React.Component {
     this.props.setParentState('residence', newData.toJS())
   }
 
+  handleClearOnConditionalChange (key, hiddenKey, value, hiddenDefaultValue) {
+    if (value === 'false') {
+    // if value is false, we want to clear the data
+    // for the field the conditonal question hides
+      let newData = Immutable.fromJS(this.props.residence)
+      newData = newData.set(key, value)
```
Nasir2712: Turns out you can use chain methods here. So instead of setting this in two different lines you could do like this. 
```newData = newData.set(key, value).set(hiddenKey, hiddenDefaultValue)```

Comment (lauherkcambria) [reply app/javascript/rfa_forms/rfa01a_edit_view/residenceCardsMain.jsx]: neat! let's keep this in mind for future iterations

File: test/javascript/rfa_forms/rfa01b_edit_view/crimeBackgroundAgainstCohabitant_tests.js [hunk 5]
```diff
@@ -52,7 +56,7 @@ describe('Verify crimeBackgroundAgainstCohabitant card', function () {
       let cardComponent = componentMountWithoutDisclosures.find('input[type="radio"]')
       let trueComponent = cardComponent.find('#crimeBackgroundAgainstCohabitantRadiotrue')
       trueComponent.simulate('change', {target: {checked: false}})
-      expect(setParentStateSpy).toHaveBeenCalledWith('arrested_for_crime', false)
+      expect(handleClearOnConditionalChangeSpy).toHaveBeenCalledWith('arrested_for_crime', undefined, 'arrested_for_crime_disclosures', [ Object({ offense: '', offense_city: '', offense_date: '', when_offense_happen: '', offense_details: '' }) ])
```
Nasir2712: any reason why this is this undefined?

PR #294, opened 2018-01-04
User county as default in facility search

sunilb0575:
-https://osi-cwds.atlassian.net/browse/CAL1-2

-User's logged in county as default in the county drop down. 

-NO

-NO

-NO


app/javascript/search/search_input.js
```diff
@@ -3,12 +3,16 @@ import InputDataBlock from './common/inputDataBlock.js'
 import {InputComponent} from 'components/common/inputFields'
 import {dictionaryNilSelectValue} from 'helpers/commonHelper.jsx'
 
+const userDefaults = Object.freeze({
+  county_name: ''
+})
+
```
lauherkcambria: what do you think about putting this in the Defaults file and import it instead?

ratneshraval:
I didn't see any new test cases. Can you add?

PR #295, opened 2018-01-08
Yes no boolean refactor

ratneshraval:
- :(

- RFA 01 C updated with YesNoComponent to use true boolean 
- RFA 01 A address card also updated to reflect how we can use boolean there.

### DIFF
--- a/app/javascript/rfa_forms/rfa01a_edit_view/addressCard.js
+++ b/app/javascript/rfa_forms/rfa01a_edit_view/addressCard.js
@@ -62,18 +62,6 @@ export default class AddressCard extends React.Component {
     this.props.setParentState('addresses', data.toJS())
   }
 
-  handleClearOnConditionalChange (typeString, key, value, hiddenKey, hiddenDefaultValue) {
-    if (value === false) {
-      let addressData = this.checkAddressType(typeString)
-      let data = addressData.data
-      data = data.update(addressData.index, x => x.set(key, value))
-      data = data.update(addressData.index, x => x.set(hiddenKey, hiddenDefaultValue))
-      this.props.setParentState('addresses', data.toJS())
-    } else {
-      this.props.setParentState(key, value)
-    }
-  }
-
   onSelection (autoFillData, typeString) {
     let addressData = this.checkAddressType(typeString)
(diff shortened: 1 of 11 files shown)


app/javascript/rfa_forms/rfa01a_edit_view/addressCard.js
```diff
@@ -62,6 +62,18 @@ export default class AddressCard extends React.Component {
     this.props.setParentState('addresses', data.toJS())
   }
 
+  handleClearOnConditionalChange (typeString, key, value, hiddenKey, hiddenDefaultValue) {
```
lauherkcambria: do we need this method here?

ratneshraval: No, that got in from a merge. I'll remove it.

app/javascript/rfa_forms/rfa01a_edit_view/addressCard.js
```diff
@@ -94,10 +94,12 @@ export default class AddressCard extends React.Component {
...
                 idPrefix='mailing_similar'
-                value={mailingAddress}
-                onFieldChange={(event) => this.props.handleClearOnConditionalChange('physical_mailing_similar', 'addresses', event.target.value, [blankMailingAddress])} />
+                value={this.props.physicalMailingSimilar}
+                onFieldChange={(selectedValue) => this.props.handleClearOnConditionalChange('physical_mailing_similar', 'addresses', selectedValue, [blankMailingAddress])}
+              />
             </div>
-            <div className={hiddenMailingSameAsPhysical}>
+
+            {this.props.physicalMailingSimilar && !this.props.physicalMailingSimilar &&
```
lauherkcambria: i'm a little confused by the logic on this line, can you explain?

ratneshraval: No worries. The logic is simply
`this.props.physicalMailingSimilar && !this.props.physicalMailingSimilar` :)
Did that help?

ratneshraval: The first part is to handle `undefined` and the 2nd part is to handle boolean part?

PR #296, opened 2018-01-09
CAL1-9 Facility-profile-Alternative phone number

sunilb0575:
-https://osi-cwds.atlassian.net/browse/CAL1-9

-Check for alternative phone number in facility profile and facility search. 

-No

-No

-No


app/javascript/search/common/commonUtils.js
```diff
@@ -16,16 +17,26 @@ export const stringForCityStateZip = (address) => {
   return address.address.city + ',' + ' ' + address.address.state + ' ' + address.address.zip_code
 }
 
-export const phoneNumberOrNa = (phone) => {
-  if (phone.length > 0) {
-    return formatPhoneNumberForDashes(phone[0])
+export const getPrimaryNumberOrNA = (phones) => {
```
lauherkcambria: these methods look like they be combined into one  by taking in the relation parameter (i.e. 'primary', 'residential', or 'Alternative'). what do you think?

ratneshraval: What Keith said plus same can be applied to address function.

ratneshraval:
@sunilb0575 I didn't see any new test cases, can you add?

app/javascript/search/search_list.js
```diff
@@ -1,28 +1,28 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import {urlPrefixHelper} from '../helpers/url_prefix_helper.js.erb'
-import {checkForNA, checkValueForNull, addressStringValueOrNa, phoneNumberOrNa} from './common/commonUtils'
+import {checkForNA, checkValueForNull, respectiveFullAddressOrNA, respectiveNumberOrNA} from './common/commonUtils'
 
 export default class SearchList extends React.Component {
   render () {
     const searchResult = this.props.searchResults
+    const primaryPhoneRelation = 'primary'
```
ratneshraval: These constants are now in 3 places, so they need to move to a common location.

ratneshraval:
@sunilb0575 I'll approve and merge now so we can verify in integration but those constants need to be extracted out and you can use same story number in new PR.

PR #292, opened 2018-01-03
Lis cws facility

Nasir2712:
#### Which User story does this relate to?
- https://osi-cwds.atlassian.net/browse/CAL1-3

#### Brief feature description
- Check to see if facility source is loaded in both grid view and list view if it exists.

#### Does this PR block/effect other stories?
- no

#### Do any commands needs to be run in order to run the project with this PR?
- no

#### Code coverage percent diff as a result of this PR?
-

PR #298, opened 2018-01-09
CAL1-3 Facility Search Results Showing LIS and CWS Facility as the Source

Nasir2712:
- https://osi-cwds.atlassian.net/browse/CAL1-3

- this PR is a bug fix for this PR - https://github.com/ca-cwds/CALS/pull/292

- no

- no

- no


ratneshraval:
Please update PR name to include story number.

Nasir2712:
@ratneshraval updated

app/javascript/search/common/commonUtils.js
```diff
@@ -43,3 +43,7 @@ export const fullName = (object) => {
 export const checkValueForNull = (value) => {
   return value == null ? 'N/A' : value
 }
+
+export const checkValueForNullOrEmptyString = (value) => {
```
ratneshraval: So why is this needed over `checkValueForNull` ? If data is empty string, it should show empty string instead of 'N/A', right?

Nasir2712: ok, i'll update it.

PR #299, opened 2018-01-09
CAL1-4 Facility Profile View: Capacity

Nasir2712:
- https://osi-cwds.atlassian.net/browse/CAL1-4

-  The capacity_last_changed is going to be the new field which has not been added yet by the api. (Will update the description once they send us this field)

- no

- n/A

-


app/javascript/search/common/commonUtils.js
```diff
@@ -24,6 +24,10 @@ export const phoneNumberOrNa = (phone) => {
   }
 }
 
+export const checkforDateOrNa = (date) => {
+  return (date === null || date === undefined) ? 'N/A' : date.split(' ')[0]
```
ratneshraval: Should do a length check too? Maybe an improvement of the following?
```
if (Boolean(date) && (date.split(' ').length > 1) )
{
  date.split(' ')[0]
} else {
  'N/A'
}
```

Nasir2712: nice i like it updating it.

test/javascript/facility/address_tests.js
```diff
@@ -27,7 +27,7 @@ describe('Verify Facility Address', function () {
       'capacity': 10,
       'license_effective_date': '2012-10-01',
       'original_application_recieved_date': '2012-07-18',
-      'last_visit_date': '2017-04-14 00:00:00',
+      'last_visit_date': '2017-04-14',
```
ratneshraval: No test for undefined/null date?

Nasir2712: added test cases for null and value

PR #300, opened 2018-01-10
CALS2-5066- Many Drop-Down Menus Not Retaining Selection - Internet Explorer

Nasir2712:
- https://osi-cwds.atlassian.net/browse/CALS2-5066

- This PR fixes these two dropdown selections in Internet Explorer
1.) "To Whom" drop down in section "IV. Minor Children Residing in the Home".
2.) "To Whom" drop down in section V. Other Adults Residing

- no

- no 

- no


spec/features/rfa_application_spec.rb
```diff
@@ -123,17 +123,22 @@
...
     expect(page).to have_content 'IV. Minor Children Residing in the Home'
     find(:select, 'relationship_to_applicant').first(:option, 'Child').select_option
+    find(:select, 'applicant_id').first(:option, applicant1FullName).select_option
     find(:select, 'child_financially_supported').first(:option, 'Yes').select_option
     find(:select, 'child_adopted').first(:option, 'Yes').select_option
     find(:select, 'minor_gender').first(:option, 'Male').select_option
     click_button('Save Progress')
     visit page.driver.current_url
     expect(find_field('relationship_to_applicant').value).to eq '1'
+    expect(find_field('applicant_id').value).to eq '168'
```
ratneshraval: Is this hard coded at 168 because of cassettes?

Nasir2712: Yes, because the applicant's id gets returned and gets stored in the value after the application is submitted.

ratneshraval: So this test won't pass in end-to-end mode. You should remove it, or read from the ID of returned response and then compare it.

Nasir2712: cool good catch! I'll try to fix this.

Nasir2712: @ratneshraval  can you take a look now?

ratneshraval: I'll retest end to end after the sprint review.

Nasir2712: sounds good to me

PR #302, opened 2018-01-10
CAL1-9 facility profile phone-number check for null

sunilb0575:
-https://osi-cwds.atlassian.net/browse/CAL1-9

-Displays N/A when primary/alternative number is null/empty. 
-Moved Alternative phone to the end of the second column. 
Extracted constants. 

-No

-No

-statements: 80, branches: 66, functions: 75, lines: 81


app/javascript/facility/address.jsx
```diff
@@ -20,31 +16,29 @@ export default class FacilityAddress extends React.Component {
               <InnerBlockAddressTitles
                 classNameTitle=''
                 title='PHYSICAL ADDRESS'
-                streetApt={respectiveStreetAddressOrNA(result.addresses, physicalAddressType)}
-                cityCountry={cityStateZipOfRespectiveAddressOrNA(result.addresses, physicalAddressType)} />
+                streetApt={respectiveStreetAddressOrNA(result.addresses, 'Residential')}
```
ratneshraval: These shouldn't be hard coded. As I mentioned in previous PR, it should be put in another file as constants to import.

app/javascript/search/common/commonUtils.js
```diff
@@ -20,7 +20,7 @@ export const cityStateZipOfRespectiveAddressOrNA = (addresses, addressType) => {
 
 export const respectiveNumberOrNA = (phones, phoneRelation) => {
   const phoneObject = phones && phones.find(o => o.relation === phoneRelation)
-  if (phoneObject) {
+  if (phoneObject && Boolean(phoneObject.number)) {
```
ratneshraval: I though Boolean was getting removed by eslint, right @Nasir2712

Nasir2712: so this will work if you have an && value to check it with because if it's just Boolean(object) then it is nothing but object for eg, if (object=null or undefined) then it is automatically false that's why you don't need Boolean to wrap it

Nasir2712:
@sunilb0575  can you add a test case with the number null in phones ?

app/javascript/facility/address.jsx
```diff
@@ -25,36 +22,34 @@ export default class FacilityAddress extends React.Component {
               <SmallInnerBlockDetails
                 classNameTitle=''
```
Nasir2712: why is classNameTitle empty?

app/javascript/search/common/commonUtils.js
```diff
@@ -30,22 +37,23 @@ export const respectiveNumberOrNA = (phones, phoneRelation) => {
 export const respectiveFullAddressOrNA = (addresses, addressType) => {
   const addressObject = addresses && addresses.find(o => o.type === addressType)
   if (addressObject) {
-    return fullAddressString(addressObject.address)
+    const streetString = checkForValueOrBlank(addressObject.address.street_address)
+    const cityString = checkForValueOrBlank(addressObject.address.city)
+    const stateString = checkForValueOrBlank(addressObject.address.state)
+    const zipString = checkForValueOrBlank(addressObject.address.zip_code)
+    const fullString = streetString + ',' + cityString + ',' + stateString + ' ' + zipString
+    return fullString !== ',, ' ? fullString : 'N/A'
```
Nasir2712: You don't have to do this. If the values are empty then you have to show them that they are empty. For eg, if city is only available then you have to show it like this ``` ,,Sacramento,```

sunilb0575: Yes, It working fine. if all the values are null that means whole address is null hence N/A will pop up, and when at least one value is present it will display that particular value. 
screen shot 2018-01-12 at 10 06 26 am
screen shot 2018-01-12 at 10 07 00 am

app/javascript/search/common/commonUtils.js
```diff
@@ -12,15 +12,22 @@ export const respectiveStreetAddressOrNA = (addresses, addressType) => {
 export const cityStateZipOfRespectiveAddressOrNA = (addresses, addressType) => {
   const addressObject = addresses && addresses.find(o => o.type === addressType)
   if (addressObject) {
-    return addressObject.address.city + ',' + ' ' + addressObject.address.state + ' ' + addressObject.address.zip_code
+    const cityString = checkForValueOrBlank(addressObject.address.city)
+    const stateString = checkForValueOrBlank(addressObject.address.state)
```
ratneshraval: so instead of doing this, one item at a time, why don't you do this,
```

export const fullAddressString = (address) => {
  return checkForValueOrBlank(address.street_address) + ',' + checkForValueOrBlank(address.city) + ',' + checkForValueOrBlank(address.state) + ' ' + checkForValueOrBlank(address.zip_code)
}
```

app/javascript/search/common/commonUtils.js
```diff
@@ -12,15 +12,24 @@ export const respectiveStreetAddressOrNA = (addresses, addressType) => {
...
     return 'N/A'
   }
 }
 
+export const stringForCityState = (address) => {
+  const citySateString = checkForValueOrBlank(address.city) + ',' + checkForValueOrBlank(address.state) + ' ' + checkForValueOrBlank(address.zip_code)
+  return citySateString === ', ' ? '' : citySateString
+}
+
+export const checkForValueOrBlank = (addressValue) => {
```
ratneshraval: The input argument should just be `value`.

app/javascript/search/common/commonUtils.js
```diff
@@ -12,15 +12,24 @@ export const respectiveStreetAddressOrNA = (addresses, addressType) => {
...
   const addressObject = addresses && addresses.find(o => o.type === addressType)
   if (addressObject) {
-    return addressObject.address.city + ',' + ' ' + addressObject.address.state + ' ' + addressObject.address.zip_code
+    return stringForCityState(addressObject.address)
   } else {
     return 'N/A'
   }
 }
 
+export const stringForCityState = (address) => {
```
ratneshraval: Is this used anymore? If not, can you remove it?

app/javascript/search/common/commonUtils.js
```diff
@@ -37,15 +46,16 @@ export const respectiveFullAddressOrNA = (addresses, addressType) => {
 }
 
 export const fullAddressString = (address) => {
-  return address.street_address + ',' + address.city + ',' + address.state + ' ' + address.zip_code
+  const stringForFullAddress = checkForValueOrBlank(address.street_address) + ',' + checkForValueOrBlank(address.city) + ',' + checkForValueOrBlank(address.state) + ' ' + checkForValueOrBlank(address.zip_code)
+  return stringForFullAddress === ',, ' ? '' : stringForFullAddress
```
ratneshraval: Good catch here

app/javascript/search/common/commonUtils.js
```diff
@@ -63,3 +73,11 @@ export const fullName = (object) => {
 export const checkValueForNull = (value) => {
   return value == null ? 'N/A' : value
 }
+
+export const primaryPhoneRelation = 'primary'
```
ratneshraval: Why word relation in `primaryPhoneRelation`?

app/javascript/search/common/commonUtils.js
sunilb0575: Yes, This method (stringForCityState) returns a string for city state and zip code.

app/javascript/search/common/commonUtils.js
sunilb0575: ok I will change it.

PR #304, opened 2018-01-12
Phone Card field bug fix - CALS-5423

nasar-cambria:
-https://osi-cwds.atlassian.net/browse/CALS-5423?filter=-1

- Phone Card Not Working Properly When Re-Entering RFA-01A

- no

- no

- N/A


app/javascript/rfa_forms/rfa01a_edit_view/applicantCard.jsx
```diff
@@ -86,7 +87,7 @@ export default class ApplicantCard extends React.Component {
             idPrefix={idPrefix}
             phoneTypes={this.props.phoneTypes}
             applicant_id={this.props.index}
-            phones={this.props.applicantFields.phones || undefined}
+            phones={checkArrayObjectPresence(this.props.applicantFields.phones) || undefined}
```
ratneshraval: This bug should be corrected in API, when there is no phone number, it should return `null` not `[]`

Nasir2712: This bug would not have been logged if the api was sending a null value to begin with. So, now if they fix it, there is no need for us to wrap the props in that method.

nasar-cambria: Agreed - That was my first approach but after discussing with Keith @lauherkcambria we decided to take this approach.

Nasir2712: also, does checkArrayObjectPresence handle null presence? if the api gets fixed in future?

nasar-cambria: I think it does. We can verify it inside ```commonHelper``` from where ```checkArrayObjectPresence``` is coming.

ratneshraval:
Can you update PR title to include story number?

nasar-cambria:
Updated story number.

lauherkcambria:
@ratneshraval  @Nasir2712 we worked on this branch together with back end and came to agree together upon this solution. Lets get this merged please.

PR #307, opened 2018-01-18
CALS-5113 - Facility profile error - No FacilityID present

Nasir2712:
- https://osi-cwds.atlassian.net/browse/CALS-5113

- Handle facility Id null value.

- no

- n/a

-


ratneshraval:
So you can use this as the story https://osi-cwds.atlassian.net/browse/CALS-5113.
I think ID needs to be used only for link/url, everywhere else you probably need to keep the license_number. We need to discuss tomorrow with PO/Business.

Nasir2712:
So ID needs to get be replaced with license number because that is the only field that is common for both LIS and CWS/CMS. If you want to use license number then we need have to more checks to see the source first and based on it assign a value.

Nasir2712:
@ratneshraval Ok updated.

Nasir2712:
checkforNA actually works for every field. The only field that's different is this district name which sends a name and a number. If the backend can just convert that to id and value, we don't have to write any other methods.

sunilb0575:
I agree, except full name, date and facility Id we can use checkForNA for all.

PR #310, opened 2018-01-18
Adding active bar to fix focus shift issue CALS-5454

nasar-cambria:
- https://osi-cwds.atlassian.net/browse/CALS-5454

- USER EXPERIENCE ISSUE:
When a card is selected, all controls on that card shift to the right. The shift appears to be in accordance to the blue "Active Card" indicator that sits on the left hand side of the card.

EXPECTATIONS:
The controls on the page should never shift for this type of visual change on the card.

- no

- no

- no


nasar-cambria:
Thanks, There are only a few places in c and packet list which is bringing in styles so instead of creating a separate stylesheet I thought of importing b stylesheet. I think once wood-duck gets updated it would be v simple to remove these imports. :)

nasar-cambria:
I like the second approach. Thanks!

PR #314, opened 2018-01-22
Relationship text bug cals 5478

lauherkcambria:
-https://osi-cwds.atlassian.net/browse/CALS-5478

- fix for multiple applicants and multiple children for text entry into the relationship freeform field.

- no

-no 

-no


ratneshraval:
So there is no gap between repeating child cards, It is really hard to know where a card starts and where it ends. Please check design.

lauherkcambria:
@ratneshraval that is per the original design. Regardless, It's not related to this bug and shouldn't hold up the approval of this PR, right?

lauherkcambria:
image
@ratneshraval as per design

lauherkcambria:
@ratneshraval agreed 👍

PR #318, opened 2018-01-23
fix end to end and feature test cases

Nasir2712:
- Fixing end to end and feature test cases so that build can pass.

-

-

-

-


ratneshraval:
Ok, FYI, this may fail as acceptance tests since just testing features means the coverage won't be 90%.

PR #323, opened 2018-01-25
CALS-5512 Coverage check excluded for feature only tests

ratneshraval:
- https://osi-cwds.atlassian.net/browse/CALS-5512

- When running just feature tests, coverage check is disabled.


package.json
```diff
@@ -60,11 +60,11 @@
...
     "all-tests": "./bin/karma;  npm run spec;",
-    "spec-end2end-features": "TEST_END_TO_END=true SELENIUM_BROWSER=HEADLESS_CHROME rspec spec/features",
-    "spec-end2end-features-chrome": "TEST_END_TO_END=true SELENIUM_BROWSER=CHROME rspec spec/features",
-    "spec-end2end-features-ie": "TEST_END_TO_END=true SELENIUM_BROWSER=INTERNET_EXPLORER rspec spec/features",
+    "spec-end2end-features": "TEST_END_TO_END=true SELENIUM_BROWSER=HEADLESS_CHROME bundle exec rspec spec/features",
+    "spec-end2end-features-chrome": "TEST_END_TO_END=true SELENIUM_BROWSER=CHROME bundle exec rspec spec/features",
+    "spec-end2end-features-ie": "TEST_END_TO_END=true SELENIUM_BROWSER=INTERNET_EXPLORER bundle exec rspec spec/features",
     "postinstall": "npm rebuild node-sass",
-    "rebuild-assets": "rake tmp:cache:clear; rm -rf public/assets public/packs public/packs-test; rails assets:precompile RAILS_ENV=test"
+    "rebuild-assets": "rake tmp:cache:clear; rm -rf public/assets public/packs public/packs-test; bundle exec rails assets:precompile RAILS_ENV=test"
```
Nasir2712: why do we need bundle exec here?

ratneshraval: We should use bundle exec for all rails commands. It makes sure that the command is executed against the context of project bundler/gems and not against system.

app/javascript/rfa_forms/rfa01a_edit_view/referencesCard.jsx
```diff
@@ -45,41 +45,41 @@ export default class ReferencesCard extends React.Component {
...
+            errors={fieldErrorsAsImmutableSet(this.props.errors.phone_number)}
+            onChange={(event) => this.props.setParentState('phone_number', event.target.rawValue, this.props.index)}
+            onBlur={(event) => this.props.validator.validateFieldSetErrorState(phoneNumberId, event.target.rawValue)} />
 
-        <InputComponent gridClassName='col-md-4' id='email'
-          value={this.props.reference.email}
-          label='Email (optional)' placeholder=''
-          type='text' onChange={(event) => this.props.setParentState('email',
-            event.target.value, this.props.index)} />
-          </div>
```
sunilb0575: I don't know I too got this same lint changes.

PR #324, opened 2018-01-25
CALS-5486 Combo Boxes are not Highlighted When Tabbing Through RFA Application

Nasir2712:
-https://osi-cwds.atlassian.net/browse/CALS-5486

- Tabbing should indicate where the user is on the form including dropdown.

- no

- no

- n/a


app/javascript/rfa_forms/rfa01a_edit_view/referencesCard.jsx
```diff
@@ -45,41 +45,41 @@ export default class ReferencesCard extends React.Component {
           suffixTypes={this.props.suffixTypes}
           prefixTypes={this.props.prefixTypes}
```
Nasir2712: This file was throwing a warning in console and karma because of the reserve word ```class```, changed that to ```className```.

Nasir2712:
jenkins, retest this please

Nasir2712:
jenkins, retest this please

app/javascript/rfa_forms/rfa01a_edit_view/referencesCard.jsx
nasar-cambria: I thought this was already updated by @sunilb0575

app/assets/stylesheets/application.scss
```diff
@@ -11,6 +11,11 @@
   border-left: 0.7rem solid transparent;
 }
 
+select:focus {
+  border: 0.1rem solid;
+  border-radius: 0.2rem;
+  border-color: #2974a2; }
```
nasar-cambria: How is this not caught by lint ?

ratneshraval: eslint only looks at JS/JSX

ratneshraval: 2 questions.
1. Shouldn't this be coming from react-select component itself?
2. If not coming from react-select, shouldn't this be set in wood-duck repo?

Nasir2712: 1) react-select component does not use html select so its styles does not include any styling for html select.
2) wood-duck is not using select html element, they are using react-select for dropdown component which does not need this style. So, that's why I added the styles here because we are using select html element in our dropdown component. In future, when we use wood-duck's dropdown component, we can just get ride of this style.

app/javascript/rfa_forms/rfa01a_edit_view/referencesCard.jsx
Nasir2712: Yes, but that PR might take a while to merge, so that's why I included that change here.

app/assets/stylesheets/application.scss
ratneshraval: Thanks for clarifying.

ratneshraval:
@Nasir2712 story number has space in it. Please correct.

PR #326, opened 2018-01-31
CALS-5531 Using Backspace After Selection in Drop Down Menu Returns User to Previous Screen

Nasir2712:
- https://osi-cwds.atlassian.net/browse/CALS-5531

- On IE, pressing backspace was navigating to the previous screen. This PR is suppose to stop this behaviour in IE.

- no

- no

- n/a


app/javascript/packs/application.js
```diff
@@ -16,3 +16,14 @@ console.log('Hello World from CALS')
 var componentRequireContext = require.context('.', true)
 var ReactRailsUJS = require('react_ujs')
 ReactRailsUJS.useContext(componentRequireContext)
+
+// Preventing event default on backspace navigation on IE
+document.addEventListener('keydown', keyDownSelect)
+function keyDownSelect (e) {
```
ratneshraval: :( Is there any other way than adding a document level event listener ?

Nasir2712: I can't think of any, Any suggestions? and also is there an issue to add this at the document level?

lauherkcambria: yikes, this is a hard one- i know switching from 'keydown' to 'keypress' will decrease calls to just non special alphanumeric key presses, but it still will be called everytime.

Really this is a browser issue that needs to be fixed by the ie11 team or the react team. Maybe we should investigate opening an issue with the react core team?

PR #332, opened 2018-02-06
cals1-5046-verify-children-section-gender

sunilb0575:
-https://osi-cwds.atlassian.net/browse/CALS-5046

-Verified Children section is displaying gender.
-Added feature test to validate gender in children section. 

-No

-No

-No


spec/features/show_facility_test_spec.rb
```diff
@@ -12,6 +12,12 @@
     expect(page).to have_text('Lederhouse Transitions')
   end
 
+  scenario 'Click toggle result and validate facility name in list view' do
+    facilities_list
+    page.find(:css,'.line_off-icon').click
```
ratneshraval: I recommend using an ID instead of trying to match on class. This works for now but may create issues later if we have similar class added to another element.
If you use class, good idea to wrap it in `within` statement so it is more precise.

spec/features/show_facility_test_spec.rb
```diff
@@ -25,10 +31,11 @@
     expect(page).to have_text('Approved')
   end
 
-  scenario 'click into facility and show children' do
+  scenario 'click into facility and validate children section' do
     facilities_list
     click_link('Lederhouse Transitions')
     expect(page).to have_text('Takahashi')
+    expect(page).to have_xpath('//tbody/tr[1]/td[4]', :text => 'F')
```
ratneshraval: What is the reason to use xpath in place or standard finder/matchers?

sunilb0575: I thought to validate data inside any particular cell of the table.

spec/features/show_facility_test_spec.rb
```diff
@@ -12,6 +12,17 @@
     expect(page).to have_text('Lederhouse Transitions')
   end
 
+  scenario 'Click toggle result and validate facility name in list view' do
+    facilities_list
+    within(pull_right)
```
Nasir2712: can you check the syntax of within, I think you can do something like this 
```within('.pull-right') do
       //code
      end ```

PR #338, opened 2018-02-12
CALS-5033-lag when entering data

nasar-cambria:
- https://osi-cwds.atlassian.net/browse/CALS-5033

- When typing data into the RFA 01A, there is a keystroke lag due to API Call on each key stroke and re rendering of the components. 

- Doesn't block any stories but it should improve application performance using Pure Components.

- To test the performance you either have to use Chrome performance tools follow the link below:
NOTE: This might not show us huge difference since this is a small change inside of nameCard Field.
https://developers.google.com/web/tools/chrome-devtools/evaluate-performance/reference

or 

2. Manual Performance Test using Rails:

Use this below timestamp and wrap the name field inside spec test:

puts Process.clock_gettime(Process::CLOCK_MONOTONIC)

//…NameField - CHANGE Faker::Name.first_name to some Random name so that we get a consistent 
value when running spec test. 

puts Process.clock_gettime(Process::CLOCK_MONOTONIC)
Byebug

Place a bye bug in the end

Target your spec test in the package.json file for a quick result. 
Ex: "spec-ci": "SELENIUM_BROWSER=CHROME bundle exec rspec spec/features/rfa_application_spec.rb:18;"

Run - npm run rebuild-assets
Then Run: npm run spec-ci

You will see the result in your console with a time interval for the test which is targeted. Compare the test using a pure component and without one but MAKE SURE TO Run - ```npm run rebuild-assets``` before running ```npm run spec-ci``` and every time we are making changes in the component. 

- NO

#### Performance difference as a result of changing a component to Pure Component.

TEST RESULT 

Without the pure component

431679.3812  
431676.869029

431676.869029 - 431679.3812 = ```2.512171```

= ```2.512171```

With Pure Component

431871.112402
431872.891662

= 1.77926  Test1
= 1.778224 Test2
= 1.812982 Test 3
= 1.743047 test 4

APROX performance difference: 2.512171 -  1.77926 = ```0.732911 MS```


ratneshraval:
@nasar-cambria can you convert nameCard.js to pure and find the comparison?

lauherkcambria:
how many tests were taken of of regular comp rendering and pure comp rendering? I'm wondering if it's possible to put this benchmark in a test and loop it 1000 times on each component to get a true reflection of avg rendering times

nasar-cambria:
@ratneshraval I tried that. It didn't show much of a difference so I removed it.

@lauherkcambria - Plenty of tests I haven't counted but just put a result of an average result I got.

package.json
```diff
@@ -39,6 +39,7 @@
     "react-widgets-moment": "^4.0.4",
     "react-wood-duck": "0.1.46",
     "react_ujs": "^2.2.0",
+    "recompose": "^0.26.0",
```
Nasir2712: do we still need this?

test/javascript/rfa_forms/rfa01a_edit_view/completeNameField_tests.js
```diff
@@ -17,7 +17,7 @@ describe('Verify Complete Name Field Component', () => {
       }
     }
     onChangeSpy = jasmine.createSpy('onChange')
-    component = shallow(<CompleteNameField
+    component = mount(<CompleteNameField
```
Nasir2712: why were this changed to mount?

nasar-cambria: Since we are using pureComponent which has an in build method ```shouldComponentUpdate``` and mount helps to call these methods.

package.json
nasar-cambria: Will leave it there for now. Incase we need to make use of recompose we don't have to re install the package.

PR #341, opened 2018-02-14
CALS-5018: Delayed geoservice call

DmitriyRudenko:
https://osi-cwds.atlassian.net/browse/CALS-5018

geo service rest call delay

NO

NO

karma code coverage: 81.32% Statements 1084/1333 67.63% Branches 280/414 76.94% Functions 437/568 82.18% Lines 1061/1291


app/javascript/components/rfa_forms/addressComponent.js
```diff
@@ -28,18 +29,25 @@ export default class AddressComponent extends React.Component {
...
-        suggestions: response
-      })
-    }).catch(() => {
-      return this.setState({
-        suggestions: []
+    let current = this
+    if (this.state.timeout) {
+      clearTimeout(this.state.timeout)
+    }
+    let timeout = setTimeout(function () {
```
ratneshraval: The timeout doesn't belong in state. Is there another way of handling this delayed call?

test/javascript/components/rfa_forms/addressComponent_tests.js
```diff
@@ -58,6 +61,7 @@ describe('verify Reusable Address Component', () => {
       headers: { 'content-type': 'application/json' }
     })
     addressComponent.instance().onSuggestionSelected('click', {})
+    jasmine.clock().tick(1000)
```
ratneshraval: The validate call should not be restricted by the timeout, only suggestions, right?

DmitriyRudenko:
@ratneshraval 

- removed redundant timeout from unit tests.
- moved timeout variable from state to instance

app/javascript/components/rfa_forms/addressComponent.js
DmitriyRudenko: yep, moved it to the instance

test/javascript/components/rfa_forms/addressComponent_tests.js
DmitriyRudenko: yep, this call is redundant, I removed it

ratneshraval:
@DmitriyRudenko I suggest a simpler solution by using `debounce` from lodash. 
That will be simple solution rather than trying to maintain timeout and clearing it ourselves.

app/javascript/components/rfa_forms/addressComponent.js
```diff
@@ -46,19 +46,24 @@ export default class AddressComponent extends React.Component {
...
-    let params = encodeURIComponent(value)
-    fetchRequest(url, 'POST', params).then(
-      response => response.json()).then(
-      (response) => {
-        return this.setState({
+    let current = this
+    if (this.fetchTimeout) {
+      clearTimeout(this.fetchTimeout)
+    }
+    this.fetchTimeout = setTimeout(function () {
```
nasar-cambria: @DmitriyRudenko To add to @ratneshraval comment setTimeout() is not a react component and there are other ways to implement this ex: 1) what Nesh suggested 
2) Using react-timeout which is a Higher order component which already has setTimeout() providing the wrapped component with safe versions of SET and CLEAR

DmitriyRudenko:
@ratneshraval  debounce is used now

ratneshraval:
@DmitriyRudenko Thanks for making changes. I approved but a good test case would be to send 3 keys and spy to make sure only 1 call gets made.

test/javascript/components/rfa_forms/addressComponent_tests.js
```diff
@@ -42,10 +43,12 @@ describe('verify Reusable Address Component', () => {
     })
   })
   afterEach(() => {
+    jasmine.clock().uninstall()
     fetchMock.restore()
   })
   it('Api success call', () => {
     addressComponent.instance().onSuggestionsFetchRequested('home', '')
+    jasmine.clock().tick(1000)
```
nasar-cambria: As discussed, are you still going to update this test case?

PR #344, opened 2018-02-15
CALS-5603 ApplicantCard subcards to pureComponent 

ratneshraval:
- https://osi-cwds.atlassian.net/browse/CALS-5603

- Converting application state to immutable and ApplicantCard->subcards to pureComponent


app/javascript/rfa_forms/rfa01a_edit_view/index.js
```diff
@@ -42,6 +42,7 @@ export default class Rfa01EditView extends React.Component {
       const countyValue = (this.props.user && this.props.user.county_code)
       this.state.application.application_county = this.props.countyTypes.find(countyType => countyType.id === parseInt(countyValue))
     }
+    this.state.application = Immutable.fromJS(this.state.application)
```
nasar-cambria: When I am trying to save any applicant it gives me an error: ```this.state.application.toJS is not a function```

nasar-cambria:
There are few other errors in console.

app/javascript/rfa_forms/rfa01a_edit_view/applicantCard.jsx
```diff
@@ -87,11 +86,10 @@ export default class ApplicantCard extends React.Component {
             idPrefix={idPrefix}
             phoneTypes={this.props.phoneTypes}
             applicant_id={this.props.index}
-            phones={checkArrayObjectPresence(this.props.applicantFields.phones) || undefined}
+            phones={this.props.applicantFields.get('phones') || undefined}
```
lauherkcambria: what does get function give us when phones is null or undefined?

lauherkcambria:
Nice! this is a lot of stuff, i want to take a little bit more time to understand how immutable js functions work. for example, looks like we can replace the 

`this.props.applicantFields.get('phones') || undefined`

`this.props.applicantFields.get('phones',  undefined)`

which is just a little cleaner.

This also makes me wonder if there are other neat little clean tricks in the immutable js library we can leverage.

PR #345, opened 2018-02-15
CALS-5543-Relationships field not retaining value

nasar-cambria:
- https://osi-cwds.atlassian.net/browse/CALS-5543

- RFA-01A - Section III - Relationships field not retaining value

- no

- no

- no


Nasir2712:
can you just update the relationship between applicants card feature test to check the value
1) after save

PR #348, opened 2018-02-16
CALS-5584 Spike on how to handle missing attributes 

ratneshraval:
- https://osi-cwds.atlassian.net/browse/CALS-5584

- Handle missing/null values and use default value research spike. 
- This will handle null values to default value conversion at rails model level and we can have cleaner code in React.


ratneshraval:
Handling missing attributes in Rails models will create a different set of issues when the data needs to be updated. Closing this PR as this approach is not needed.

PR #349, opened 2018-02-16
CALS-5550: remove leading spaces in names

DmitriyRudenko:
https://osi-cwds.atlassian.net/browse/CALS-5550

#### Brief feature 
Per conversation with Leonid, it may still be the case that applicant information is created when NULL applicant exist on the form.

*Acceptance Criteria*
- Null applicant information will not be created until applicant is added to RFA form and saved.
- User is not able to enter blank space as first character of name with no additional alpha-numeric values  behind it.
Example: If user entered ' ' (blank space only) it will be considered no value
If user enters ' 'Leon, application will accept the leading space and save with the name.
DEV NOTE: Determine if in case number two, the leading space to start a name can be parsed out and only save as 'Leon'
- Spaces anywhere else in the name structure are permitted.

NO

NO


Nasir2712:
can you add description to PR instead of NO

Nasir2712:
Can you write new feature tests and karma tests for the component that you made changes too?

DmitriyRudenko:
@lauherkcambria 

With your code user will still able to enter leading spaces. Please review the 2nd requirement more carefully.

Also couple of remarks:

1. Multiple trims on the same value is not really efficient
2. first_name or last_name my be undefined, so better to check it
3. Why do we check just first entry here? What is the hidden meaning of "0" index? should it be replaced by constant with more understandable name?
4. I believe you did a typo here: String.length type is integer
5. What function in index.js did you mean? I didn't change any of index.js

Thanks!

LeonidMarushevskyi:
@DmitriyRudenko You need to recird VCR cassette for your test

LeonidMarushevskyi:
PR reworked according to review comments. Feature tests added

lauherkcambria:
image
hmm, something seems a little off when I run this branch from terminal. instead of the save progress button being disabled the add alias button is disabled. See pic

ratneshraval:
@DmitriyRudenko One of @Nasir2712 's request was to add Karma test for the new function to verify functionality. I didn't see that but we have 2 approvals so I"ll merge it. Can you add the karma test for it as well?

PR #350, opened 2018-02-16
CALS-5559 facility search pagination

sunilb0575:
-https://osi-cwds.atlassian.net/browse/CALS-5559

-Remove default 10 results limit and add pagination to facility search

-No

-No

-82  Statements  69 Branches  78 Functions  83 Lines


lauherkcambria:
a few questions on functionality: 
1. is there a default here for the items to be displayed? what increment(s) are we cutting off at? some description may be necessary when there are less than 5 items:
image
2. if there are less than 2 pages, it looks like the next page icon is still clickable. this makes it show zero results. 
image
3. toggling between grid and table view isn't available anymore. is this by design?
image
4. nothing collapeses/hides when the triangle left of 'Advanced Search' is clicked. is this intentional?
image
5. I don't see anywhere the search critieria being displayed once a search is performed. is this intentional? 
image

lauherkcambria:
searching without parameters gives all 
image

app/controllers/facilities_controller.rb
```diff
@@ -24,9 +24,15 @@ def search
     es_query_json = Elastic::QueryBuilder.facility_search_v1(query_hash).to_json
     logger.info "es query: #{es_query_json}"
     @facilities = facility_helper.search es_query_json
-    @facilities = @facilities['hits']['hits'].collect { |facility| facility['_source']}
-    @facilities.sort_by! {|facility_name| facility_name['name']}
-    json_response @facilities
+    # @facilities = @facilities['hits']['hits'].collect { |facility| facility['_source']}
```
ratneshraval: Please remove commented code here.

app/javascript/search/search.js
```diff
@@ -38,27 +50,84 @@ export default class Search extends React.Component {
     this.setState({isToggled: !this.state.isToggled})
   }
 
-  addSearchInput (DataSearch) {
-    var query = DataSearch.split(',')
+  addSearchInput (dataSearch, getFromValue, getSizeValue) {
+    let dataWithNoFromSize = dataSearch
+    let newDataSearch = dataSearch + ',' + getFromValue + ',' + getSizeValue
```
ratneshraval: `from` and `size` parameters should not be merged with query block. That is making your code more complex than necessary.
Consider keeping them separate and adding as query string parameter search call like `/facilities/search?from=2&size=10`

app/javascript/search/search.js
```diff
@@ -14,11 +14,23 @@ export default class Search extends React.Component {
       landingPageUrl: props.landingUrl,
       isToggled: true,
       inputData: ' ',
-      searchData: undefined
+      searchData: undefined,
+      totalNoOfResults: 0,
```
ratneshraval: One major rule to consider when adding anything to state, whatever doesn't belong in UI or can be calculated from other props, shouldn't be added to state.

app/javascript/search/search.js
```diff
@@ -38,27 +50,84 @@ export default class Search extends React.Component {
...
       id: [query[2]],
       name: [query[3]],
-      'addresses.address.street_address': addressData
+      'addresses.address.street_address': addressData,
+      'from': [getFromValue],
+      'size': [getSizeValue]
     }
+    // call http request function with arguments
+    let url = '/facilities/search'
+    fetchRequest(url, 'POST', params).then((response) => {
```
ratneshraval: There is already a `POST facilities/search` (line 122), we shouldn't have 2. Consider refactoring.

app/javascript/search/search_Data.js
```diff
@@ -42,8 +44,20 @@ export default class SearchDetails extends React.Component {
     this.props.sendSearchInput(newQuery)
   }
 
+  arrayBuilder (searchCount) {
+    let arrayResult = []
+    for (let i = 1; i <= searchCount; i++) {
+      if (i % 5 === 0) {
+        arrayResult.push(i)
```
ratneshraval: I like the dynamic approach but recommend cutting this off at 50.

sunilb0575:
@lauherkcambria 
For question 
1. As we don't have the specific design for limiting the values in drop down, for this PR the dropdown values are generated based on the total number of results. 
2. Good catch, Fixed it. 
3. Done.
4. Again as we don't have design for the advance search, leaving it as is.
5. Right now the search criteria is displayed only when search by facility Id and facility Name.

screen shot 2018-02-21 at 2 27 20 pm

sunilb0575:
@ratneshraval I can address  rest of the comments, while refactoring.

app/javascript/search/search.js
```diff
@@ -14,79 +14,126 @@ export default class Search extends React.Component {
...
+        inputData: DataSearch,
+        searchResults: data.facilities,
+        totalNoOfResults: data.total,
+        disableNext: false
       })
-      .catch(error => {
-        console.log(error)
-        return this.setState({
-          searchData: []
+      if (this.state.fromValue - this.state.sizeValue < 0 || getFromValue === 0) {
```
ratneshraval: These 2 should be a function. `nextPageLinkStatus`, `previousPageLinkStatus`
When you have those, you can have a single `setState` call and inside you can do
```
this.setState({
        inputData: DataSearch,
        searchResults: data.facilities,
        totalNoOfResults: data.total,
        disableNext: nextPageLinkStatus(page, total, ....)
        disablePrevious: previousPageLinkStatus(page, total, ...),
        fromValue: getFromValue,
        sizeValue: getSizeValue,
        pageNumber: 1
      })
```

app/javascript/search/search_Data.js
```diff
@@ -73,14 +68,26 @@ export default class SearchDetails extends React.Component {
           <span>Advanced Search</span>
         </div>
         <div className='search_details col-xs-12 col-sm-9 col-md-9 col-lg-9'>
-          <p>Search Results: <span>1-{searchCount.length}</span></p>
+          <p>Search Results:</p>
+          <select
+            className='search_dropdown'
+            id='dropdown_no_of_facilities'
```
ratneshraval: why snake_casing for these names?

app/javascript/search/search.js
```diff
@@ -14,79 +14,135 @@ export default class Search extends React.Component {
...
+  handleChange (facilitiesPerPage) {
+    this.setState({
+      sizeValue: parseInt(facilitiesPerPage),
+      fromValue: 0
+    }, () => {
+      this.searchApiCall(this.state.inputData, this.state.fromValue, this.state.sizeValue)
+    })
+  }
+
+  nextPageLinkStatus (total) {
```
ratneshraval: Can you make this function not use state? Pass fromValue as an argument.

app/javascript/search/search.js
```diff
@@ -14,79 +14,135 @@ export default class Search extends React.Component {
...
+    if (this.state.fromValue + this.state.sizeValue >= total || total < 5) {
+      return true
+    }
+  }
+
+  previousPageLinkStatus (fromValue, sizeValue) {
+    if (this.state.fromValue - this.state.sizeValue < 0 || fromValue === 0) {
```
ratneshraval: Same stateless comment here.

spec/classes/elastic/query_builder_spec.rb
```diff
@@ -32,20 +32,18 @@
...
       }
 
-      input = [ {fac_co_nbr: '28', fac_name: 'home'}, {fac_co_nbr: '18'} ]
-      output = Elastic::QueryBuilder.match_boolean(input)
-
+      input = [{"fac_co_nbr"=>"28", "fac_name"=>"home"}]
+      from = "0"
+      size = "5"
+      output = Elastic::QueryBuilder.match_boolean(input, from, size)
+      puts output
```
Nasir2712: remove this puts

PR #305, opened 2018-01-15
CALS1-80 upgrade to perry v2

Nasir2712:
#### Which User story does this relate to?
- https://osi-cwds.atlassian.net/browse/CALS-5071

#### Brief feature description
- Upgrading to Perry v2 to stay updated to the newest architecture. 
Perry - Browser -> Token -> Token Validation -> Login
Perry V2 - Browser -> Access Code -> Token Generate -> Token Validation -> Login

Fixes facility both end to end and feature tests.

#### Does this PR block/effect other stories?
- Yes PR's might fail until this PR goes in first because I modified env variables in Jenkinsfile.

#### Do any commands needs to be run in order to run the project with this PR?
**change you're env variables** -
CALS_API_URL=https://calsapi.preint02.cwds.io
BASE_SEARCH_API_URL=https://doraapi.preint02.cwds.io
AUTHENTICATION_API_BASE_URL=https://web.preint02.cwds.io/perry

Modify the TOKEN variable from the terminal.

Testing end to end on Chrome - **yarn run spec-end2end-features-chrome**

#### Code coverage percent diff as a result of this PR?
- na

PR #352, opened 2018-02-23
Upgrade to Perry V2

Nasir2712:
- Related to this https://github.com/ca-cwds/CALS/pull/305

- Upgraded to PerryV2. Verified and Tested

-

-

-


Nasir2712:
@ratneshraval If everything looks ok to you. I will close the other PR.
Thanks!

app/controllers/cals_base_controller.rb
```diff
@@ -42,16 +42,17 @@ def set_relationship_to_applicants(parameters, applicants)
   end
 
   def authenticate_with_cwds
-    session.delete(:token) if params[:token].present?
+    token = Cwds::Authentication.token_generation(params[:accessCode], AUTHENTICATION_API_BASE_URL) if session[:user_details].blank?
```
lauherkcambria: there are a lot of if statements here. anyway we could refactor this to reduce complexity?

PR #354, opened 2018-02-26
Disable save bug fix for cals 5619

lauherkcambria:
- https://osi-cwds.atlassian.net/browse/CALS-5619

-Save Progress disabled when both Legal First Name and Legal Last Name are missing
Save Progress disabled when Legal First Name is present but Legal Last Name is mssing
Save Progress disabled when Legal First Name is missing but Legal Last Name is presented

- n/a

- n/a

-
81.77% Statements 
 68.1% Branches
 77.72% Functions 
82.4% Lines


lauherkcambria:
note: all dev merges need 2 reviewer approvals before merge.

PR #355, opened 2018-02-27
CALS-5617 Upgrade webpacker to 3.2

ratneshraval:
- https://osi-cwds.atlassian.net/browse/CALS-5617

- Upgrade webpacker to 3.2

- Need it for production build story

- `bundle install`
- `yarn install`


ratneshraval:
@lauherkcambria the PR is updating webpacker to 3.2. Unfortunately there isn't a good documentation for upgrade. I upgraded and went through all config changes comparison between current and new version.
If you see, all of those loaders are now gone from our project and handled by webpacker itself. We can customize it if we want, like how I'm doing in `loaders/file` and `loaders/erb`.

Nasir2712:
@ratneshraval I see some ruby lint errors. Can we fix them? Also, does rubocop provide a way to run lint as a script?

ratneshraval:
@Nasir2712 the lint errors are rubocop issue. If I fix it, it will not be consistent with how we have them in other places. For example space after `{`.
There is an auto fix option. I'll create a separate PR since it says 1000 files changed :)

PR #356, opened 2018-02-27
CALS-5540 DEV - RFA Side Nav Reflects Scrolling or Focus

lauherkcambria:
-https://osi-cwds.atlassian.net/browse/CALS-5540

implemented scrollspy so that the user can see where on page they are in the sidebar when scrolling in the page. added feature tests.

- no

- npm run rebuild-assets

n/a, added rspec feature tests.


spec/support/capybara.rb
```diff
@@ -46,7 +46,7 @@ def register_internet_explorer(capybara_config, selenium_browser)
 end
 
 def registar_chrome(capybara_config, selenium_browser)
-  capability_options = selenium_browser == :headless_chrome ? %w[headless disable-gpu no-sandbox] : []
+  capability_options = selenium_browser == :headless_chrome ? %w[headless disable-gpu no-sandbox window-size=1200,719] : []
```
ratneshraval: Why was the window size added?

package.json
```diff
@@ -34,7 +34,9 @@
     "react-overlays": "^0.8.3",
     "react-redux": "^5.0.3",
     "react-router": "^3.0.2",
+    "react-scrollspy": "^3.3.5",
```
ratneshraval: Is react-scrollspy still a dependency?

app/javascript/rfa_forms/rfa01a_edit_view/index.js
```diff
@@ -156,150 +157,190 @@ export default class Rfa01EditView extends React.Component {
...
-            stateTypes={this.props.stateTypes}
-            references={(this.state.application.get('references') && this.state.application.get('references').toJS()) || undefined}
-            suffixTypes={this.props.suffixTypes}
-            prefixTypes={this.props.prefixTypes}
-            nameTypes={this.props.nameTypes}
-            validator={this.validator}
-            errors={this.state.errors.reference} />
-        </div>
-
+        <Waypoint
```
ratneshraval: Can we have waypoint refactored so the common props `scrollableAncestor={window} bottomOffset='90%'` only live in 1 place?

spec/support/capybara.rb
lauherkcambria: this was added in order for headless chrome to be able to execute the javascript 'window.scrollTo' function in order to test scroll spy functionality. This size was chosen because it's the default size the browser pops up as when we run it with CHROME instead of HEADLESS_CHROME from terminal. This line will allow us to further fully test feature positioning.

package.json
lauherkcambria: no - i'll remove it.

app/javascript/rfa_forms/rfa01a_edit_view/index.js
lauherkcambria: what do you suggest?

lauherkcambria:
@ratneshraval unfortunately as discussed, that refactor doesn't work because of the explanation available at https://github.com/brigade/react-waypoint/issues/159 and https://github.com/brigade/react-waypoint/issues/207.

ratneshraval:
@lauherkcambria have you tried making it a class component? I think the limitation is with the functional component only.

lauherkcambria:
@ratneshraval good idea- that looks like it will work, i switched the functional component to a class component.

app/javascript/components/common/scrollSpy.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import Waypoint from 'react-waypoint'
+
+export default class ScrollSpy extends React.PureComponent {
+  render () {
+    return (
+      <Waypoint
+        scrollableAncestor={window}
+        onEnter={this.props.onEnter}
```
ratneshraval: I think you forgot to add propTypes here.

lauherkcambria: @ratneshraval addressed.

PR #357, opened 2018-03-01
Rubocop autofix for app folder

ratneshraval:
- n/a

- rubocop autofix for app folder. I don't think it is needed but wanted to generate so everyone can see.


ratneshraval:
Closing this as basically this is auto fix is not what we want.

PR #358, opened 2018-03-01
Rubocop yml

ratneshraval:
- n/a

- Ignore line length flags


ratneshraval:
@lauherkcambria it has lot of rules but the rules don't match one on one to codeclimate. 
if you see a rule bothering in codeclimate, we can try to see how it relates in rubocop. remember, rubocop only deals with ruby and code climate deals with everything.

PR #359, opened 2018-03-01
Backspace text area fix cals 5637

lauherkcambria:
- https://osi-cwds.atlassian.net/browse/CALS-5637

- fix for backspace bug in body of water and directions to house that was intorudeced as a part of fixing the ie backspace issue. 

- no

- no

- 82.56% Statements 
 69.06% Branches 
 78.19% Functions
 83.15% Lines


Nasir2712:
can you update the test case in rspec?

nasar-cambria:
Except the above comment everything else looks good.

app/javascript/packs/application.js
```diff
@@ -20,11 +20,14 @@ ReactRailsUJS.useContext(componentRequireContext)
...
   let keyCode = e.which
-  if (document.activeElement === document.body ||
+  if ((document.activeElement === document.body ||
       document.activeElement.tagName === 'SELECT' ||
       document.activeElement.tagName === 'BUTTON' ||
       document.activeElement.tagName === 'A' ||
-      document.activeElement.type !== 'text'
+      document.activeElement.type === 'radio' ||
+      document.activeElement.type === 'checkbox') &&
+     (document.activeElement.type !== 'text' ||
```
Nasir2712: radio and checkbox are input types as well. document.activeElement.type !== 'text' actually takes care of them. I don't think you need to add a check for radio or checkbox again.

lauherkcambria: unfortunately, we need to. not adding those two lines makes the return to previous page issue in IE occur if the user hits the backspace on either a checkbox or radio. Thus, these two lines are necessary.

PR #361, opened 2018-03-02
WIP-CALS1-5027-sort facility search results by facility name

sunilb0575:
- https://osi-cwds.atlassian.net/browse/CALS-5027
-Blocked by https://osi-cwds.atlassian.net/browse/CALS-5659

-Sort facility search results.
-Sort order by facility name. 

- No

- No

- WIP


ratneshraval:
@sunilb0575 the sort field should be passed in instead of hard coding for `name`

app/classes/elastic/query_builder.rb
```diff
@@ -45,20 +45,29 @@ def self.match_boolean(query_array, from_params, size_params)
           should: combined_query_array
         }
       },
+      sort:
+      [
+        {
+          sort_by_params =>
```
ratneshraval: please use `:` instead of arrow `=>` notation

ratneshraval:
@sunilb0575 if you add a way to only add sort logic to elasticsearch query when a field is specified, then you can go ahead and merge this branch and not specify search field till backend story is done.

app/classes/elastic/query_builder.rb
sunilb0575: As *sort_by_params* is a string, I am getting an syntax error when used *:*

PR #362, opened 2018-03-02
CALS-5445: fix for employee address not able to backspace bug

lauherkcambria:
- https://osi-cwds.atlassian.net/browse/CALS-5445

- as a result of the switching some cards to immutable.js last week, some regression was introduced. this fixes that bug. 

- 

- 

-


app/javascript/rfa_forms/rfa01a_edit_view/applicantCard.jsx
```diff
@@ -72,7 +72,7 @@ export default class ApplicantCard extends React.Component {
           <Employment
             stateTypes={this.props.stateTypes}
             salaryTypes={this.props.salaryTypes}
-            employment={this.props.applicantFields.get('employment') || undefined}
+            employment={Immutable.fromJS(this.props.applicantFields.get('employment')) || undefined}
```
Nasir2712: So we are setting the state of the application in Immutable JS - then why do we have to use Immutable.fromJS again here?

lauherkcambria: because the nested fields need to be immutable.applicantFields,employment.physical_address  isn't immutable, we need to do a Immutable.fromJs() here in order to have the nested fields be immutable as well (one of the downsides of this refactor).

ratneshraval:
@lauherkcambria 
1. Can you add a karma test that fails in current situation that this code fixes?
2. The problem stems from not having address object as immutable when selecting one of the suggestions. A smaller fix would be to update line 20 in `employmentCard.jsx` to make sure set happens using immutable
```
 onEmploymentChange (key, value) {
    const newData = this.props.employment.set(key, Immutable.fromJS(value))
    this.props.setParentState('employment', newData)
  }
```

ratneshraval:
Whenever we fix a bug we should create a test that fails so the exact same bug won't come back. I cannot approve with PR without the test.
Although, I'll help create the test so we can get this in before sprint closes.

lauherkcambria:
thank you for your assistance! 😄

PR #376, opened 2018-03-13
CALS-5574 Facility search - Page header

Nasir2712:
-  https://osi-cwds.atlassian.net/browse/CALS-5574

- add breadcrumbs to the body of the page below page header in search.

- no

- no

- N/A


ratneshraval:
@Nasir2712 remove the spaces from story number

test/javascript/components/common/breadCrumb_tests.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import BreadCrumb from 'components/common/breadCrumb'
+import {shallow, mount} from 'enzyme'
+
+describe('bread crumb rendering', () => {
+  let breadCrumbComp
+  it('verify breadCrumb rendering', () => {
+    let props = <a href="/search"> Facility Search</a>
+    breadCrumbComp = mount(<BreadCrumb children={props} />)
```
ratneshraval: Do you need `mount` here vs `shallow`?

app/javascript/components/common/breadCrumb.js
```diff
@@ -0,0 +1,28 @@
...
+import {urlPrefixHelper} from 'helpers/url_prefix_helper.js.erb'
+
+const BreadCrumb = ({
+  breadCrumbSign,
+  children
+}) => (
+  <div className='container'>
+    <div className='row'>
+      <div className='col-xs-7'>
+        <h5>Back to: <span><a href={urlPrefixHelper('/')}>Main Menu</a> {breadCrumbSign} {children}</span></h5>
```
ratneshraval: What happens when you're 2 levels deep? Would we need to update this component at that time?

Nasir2712: yes we have too

PR #378, opened 2018-03-15
CALS-5568: required field indicators added to RFA 1B

LeonidMarushevskyi:
- https://osi-cwds.atlassian.net/browse/CALS-5568

- Minimum Required Indication added to RFA 1B form

- No

- No

- 82.96% Statements 
- 70.32% Branches 
- 78.8% Functions 
- 83.54% Lines


nasar-cambria:
@LeonidMarushevskyi According to the acceptance criteria the first field ```what was the offense``` is not marked required. I see you added ```(required)``` to that label is that intentional ?

Please find the A/C from the Jira story below:


Disclosure of Criminal Background - Against Child / Spouse / Cohabitant
Given I have selected YES to Have you ever been arrested for a crime against a child or for spousal/cohabitant abuse?

Where did the offense happen is marked as required
When did the offense happen is marked as required
Explain what happened is marked as required
Given that I have selected to add another offense against child...

nasar-cambria:
@vladimirPe I think it is also possible to add the content ```(required)``` in CSS for ex: ```.required:after { content:' (required)'; }``` and just add a class attribute to the field which works good with the validation also that checks the form or highlights improperly completed fields. Also, it doesn't add any additional mark up inside label element.

app/javascript/rfa_forms/rfa01b_edit_view/outOfStateDisclosureCard.js
```diff
@@ -37,7 +38,7 @@ export default class OutOfStateDisclosureCard extends React.Component {
         {livedInOtherState === 'true'
           ? <div>
             <MultiSelect
-              label={Rfa01bOutOfStateDisclosureCardText.identifyStates}
+              label={livedInOtherState ? Rfa01bOutOfStateDisclosureCardText.identifyStates + RfaCommon.requiredIndicator : Rfa01bOutOfStateDisclosureCardText.identifyStates}
```
ratneshraval: you already have `livedInOtherState === true` before you get to this line. Do you need conditional?

LeonidMarushevskyi:
@nasar-cambria Confirmed with Leon that What was the offense? must be required as well.

LeonidMarushevskyi:
@ratneshraval Unneeded conditional logic removed

ratneshraval:
@LeonidMarushevskyi I think you missed my comment about test. Can you add a jasmine test for 1 condition added?

LeonidMarushevskyi:
@ratneshraval I didn't missed your comment. Was working on tests. Tests added. Please review

test/javascript/rfa_forms/rfa01b_edit_view/applicantDetailsCard_tests.js
```diff
@@ -62,6 +95,18 @@ describe('Verify ApplicantDetailsCard card', function () {
       namePrefixTypes={prefixTypes.items}
       nameSuffixTypes={suffixTypes.items} />)
 
+    componentMountNoLicenseNumber = shallow(<ApplicantDetailsCard
```
ratneshraval: You shouldn't load 3 components on `beforeEach`. I recommend you have a separate `describe` block so only needed components are loaded.
Also if you are `shallow` rendering, use the variable name to have shallow instead of mount. (e.g `componentMount` vs `componentShallow`

LeonidMarushevskyi: @ratneshraval test applicantDetailsCard_tests.js refactored. Please review.

test/javascript/rfa_forms/rfa01b_edit_view/applicantDetailsCard_tests.js
```diff
@@ -77,45 +86,109 @@ describe('Verify ApplicantDetailsCard card', function () {
...
+        setParentState={setParentStateSpy}
+        stateTypes={stateTypes.items}
+        namePrefixTypes={prefixTypes.items}
+        nameSuffixTypes={suffixTypes.items} />)
+    })
+
+    it('check required fields indicators', () => {
+      let componentHtml = shallowMount.html()
+      expect(componentHtml).toContain('Name of Resource Family (required)')
+      expect(componentHtml).toContain('First Name (required)')
```
ratneshraval: These kind of tests do not really check if correct item has correct label. It just check that somewhere in the HTML the text is found. It is better to find the individual item by id/class and check the text/label to make sure correct item has correct text.

ratneshraval:
@LeonidMarushevskyi I approved the changes. Check my comment on html text verification.

test/javascript/rfa_forms/rfa01b_edit_view/applicantDetailsCard_tests.js
LeonidMarushevskyi: @ratneshraval Currently in application there is no simple way to locate labels. Take a look at I.Out of State Disclosure in RFA 1A and try to locate label "If YES, identify each state and complete a LIC 198B for each state listed". Label don't have any class/id and not associated with any control. So to implement your suggestion I will need to change code. Not sure that it worth it. 
I don't see possibility a component to have several similar label. Tests are testing business requirements that labels has required mark. So if someone will remove them test will show regression. I believe this tests are simple and doing what they need to do.

PR #382, opened 2018-03-16
Cals 5569 Minimum Required Indication in UI for RFA 1C

vladimirPe:
- https://osi-cwds.atlassian.net/browse/CALS-5569

- As a user, I want to know what fields are required in order to submit the RFA 01C form so that I can complete the application accurately the first time.

- No

- No

- No


test/javascript/rfa_forms/rfa01c_edit_view/desiredChildcard_tests.js
```diff
@@ -132,6 +133,14 @@ describe('Verify RFA 01C child desired', function () {
       [{applicant_id: 20, relationship_to_applicant_freeform: 'test', relationship_to_applicant: {value: '', id: 0}}])
   })
 
+  it('verify required fields labels', () => {
+    let componentHtml = childCardComp.html()
+    expect(componentHtml).toContain('Is the child currently in your home?' + RfaCommon.requiredIndicator)
```
ratneshraval: These kind of tests do not really check if correct item has correct label. It just check that somewhere in the HTML the text is found. It is better to find the individual item by id/class and check the text/label to make sure correct item has correct text.

vladimirPe: fixed for: 
- First Name
- Last Name
- Date Of Birth

but for "Is the child currently in your home?" there is no id, class or for an attribute which can be used as a unique identifier. 
If you have any idea how to identify this label please share the information

ratneshraval: That's good. At least we are isolating to only 1 component. Thanks for fixing the others.

ratneshraval:
jenkins, retest this please

PR #384, opened 2018-03-19
CALS-5561 Feature Enable/Disable Flag

Nasir2712:
-  https://osi-cwds.atlassian.net/browse/CALS-5561

-  setting an env variable which turns on/off features. This PR enables you to hide the create rfa application, hide the applications table and disable the routes to the existing applications.

- no

-  add FEATURE_DISABLE_RFA_APPLICATION=TRUE to your .env file

- n/a


Nasir2712:
if everything looks good, I'll add a new env variable to jenkins file for testing purposes.

app/controllers/welcome_controller.rb
```diff
@@ -2,12 +2,17 @@ class WelcomeController < CalsBaseController
...
     @applications = rfa_application_helper.all_expanded
+    @disable_rfa_application = disable_rfa_application
   end
 
   def create
 
   end
 
+  def disable_rfa_application
+    DISABLE_RFA_APPLICATION == 'TRUE'
```
ratneshraval: Why choice of string over boolean?

nasar-cambria:
I get this error in ie

screen shot 2018-03-19 at 1 41 47 pm

spec/features/dashboard_test_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'rails_helper'
+require 'vcr'
+require 'faker'
+
+
+RSpec.feature 'Dashboard', js: true, set_auth_header: true  do
+  before(:each) do
+    allow_any_instance_of(WelcomeController).to receive(:disable_rfa_application).and_return(true)
```
ratneshraval: Instead of method, is there a way to set value of the variable directly? So that you can avoid unnecessary method
How about `before` method where you set the value of `DISABLE_RFA_APPLICATION` to true and set it back to original in `after`

Nasir2712:
@nasar-cambria  add DISABLE_RFA_APPLICATION=TRUE to your .env file

spec/features/dashboard_test_spec.rb
Nasir2712: I tried every way possible to set it up directly without trying to wrap up in a method. Nothing worked. 
```#stub_const('ENV', {'DISABLE_RFA_APPLICATION' => 'TRUE'})
# ENV.stub(:fetch).with("DISABLE_RFA_APPLICATION").and_return("TRUE")
# allow(ENV).to receive(:[]).with(DISABLE_RFA_APPLICATION).and_return(TRUE)
#allow(ENV).to receive(DISABLE_RFA_APPLICATION).and_return("FALSE") #ENV['DISABLE_RFA_APPLICATION'] = 'TRUE'
```

app/controllers/welcome_controller.rb
Nasir2712: ya you don't need a check there. removing it

spec/features/dashboard_test_spec.rb
ratneshraval: `stub_const('DISABLE_RFA_APPLICATION', 'TRUE')`

Nasir2712: hmm let me try that

Nasir2712: great that worked!!

ratneshraval:
@Nasir2712 
the routes file can be more readable if wrapped in an `if` block.
But since the button is disabled and not gone, we can't entirely eliminate the routes.

Nasir2712:
yup that's an issue. Let's keep it as it is for now

Nasir2712:
Jenkins, retest this please

ratneshraval:
@Nasir2712 can you rename variable to include `FEATURE_` in front so it is easy to identify all feature flags.

config/initializers/1_ENV_VARS.rb
```diff
@@ -9,6 +9,12 @@
 
 SANDBOX_LANDING_URL = ENV['SANDBOX_LANDING_URL']
 
+if ENV['DISABLE_RFA_APPLICATION'] == 'TRUE'
```
ratneshraval:
```
DISABLE_RFA_APPLICATION = ENV['DISABLE_RFA_APPLICATION'] == 'TRUE'
```

Nasir2712:
@ratneshraval DISABLE_FEATURE_RFA_APPLICATION? or FEATURE_DISABLE_RFA_APPLICATION?

config/initializers/1_ENV_VARS.rb
Nasir2712: nice, neat trick!

PR #387, opened 2018-03-20
Logout url using props

ratneshraval:
- n/a

- fixing logout bug

TODO: Add a failing test.


Nasir2712:
you can comment out the rspec test case for logout. That will fail before the fix and should pass after the fix. 
```spec/features/global_header_test_spec.rb```

ratneshraval:
Trying to use the jasmine test as that will run faster. 
What I read is, we should be able to use one of the 2 and spy on it
1. window.location.href=x
2. window.location.assign(x)

ratneshraval:
Closing in favor of #434

PR #389, opened 2018-03-21
Update dockerfile

speddi01:
- 

- 

- 

- 

-


speddi01:
updated with volume

PR #391, opened 2018-03-22
DOE-2637 disable static assets from rails

ratneshraval:
- https://osi-cwds.atlassian.net/browse/DOE-2637

- disable static assets from rails server


Srikanthchowdhary:
Approve the change

PR #396, opened 2018-03-23
CALS-5809 Rspec - 2 deprecation warnings

sunilb0575:
- https://osi-cwds.atlassian.net/browse/CALS-5809

- Fix deprecation warning when Rspec spec.
-Switch from Rspec 3.5 to Rspec 3.7.
-Gem used to update syntax.
https://github.com/yujinakayama/transpec

- No

- bundle install

- No


Nasir2712:
update the instructions on how to use the new gem.

ratneshraval:
Do we need to keep the gem if all has been updated? If we need the gem, please move it to `development and test` or `test` group

sunilb0575:
I think we don't need Transpec gem in the future. It just helps in updating syntax, which I already did on whole project. 
Any ways I will keep a link related to gem in the description.

Nasir2712:
I think you should still keep it in the test group so that we could still use it incase if we need it.

Nasir2712:
also, take the screenshots so that you can show in JIRA.

PR #399, opened 2018-03-27
CALS-5735 Relationships text boxes now displaying

nasar-cambria:
- https://osi-cwds.atlassian.net/browse/CALS-5735

- When generating an RFA 01C, there were previously text boxes showing, allowing the user to type in a relationship for each child to each applicant. Today's test, there is a large header visible, but no text boxes to enter relationships.

- no

- no

- N/A


app/assets/stylesheets/application.scss
```diff
@@ -15,6 +15,21 @@
   white-space: nowrap;
 }
 
+.child-name-section {
```
ratneshraval: Why rfa 01 c specific styles in application.scss?

nasar-cambria: Keeping it at this level to make use of it in 01A and 01C as well. Those forms needs clean up too. Some of the input fields and texts are not correctly aligned.

ratneshraval: If it is needed for others then please create a new file that you can import instead of adding to application

nasar-cambria: That make sense. Done!

app/javascript/rfa_forms/rfa01c_edit_view/desiredChildDetails.js
```diff
@@ -18,7 +18,7 @@ const DesiredChildDetails = ({
   countyTypes
 }) => (
   <div className='row'>
-    <div className='card-header'>Name of the Child</div>
+    <div className='child-name-section'><span className='row-margin'>Name of the Child</span></div>
```
lauherkcambria: i like this naming convention, nice.

nasar-cambria: thanks!

app/assets/stylesheets/application.scss
```diff
@@ -6,6 +6,7 @@
 @import "font-awesome";
 @import 'react-wood-duck/dist/styles/application.scss';
 @import './sidebar.scss';
+@import './application_relationship.scss';
```
anuroop15: @nasar-cambria, any need in creating a new file for just relationship card. can this styles be moved to application.scss..?

nasar-cambria: The idea is it keep these new styles in a separate file and import it in application.scss to keep it clean.

PR #402, opened 2018-03-29
WIP CALS-5696 Removing (required) label from Employment Card

nasar-cambria:
- https://osi-cwds.atlassian.net/browse/CALS-5696

- Since the (required) label in address fields are coming from wood-duck its adding (required) label to every common address component in RFA 01A. We need to make sure this label is only added to fields which are required.

- no

- no

- NA


ratneshraval:
There is a constant declared for ` required` text. shouldn't we be using that?

Nasir2712:
I was never a fan of that approach having react component evaluating the object at rendering time for a string. Unnecessary evaluation.

ratneshraval:
@Nasir2712 That is a good point. We should've corrected/commented when that approach got it.

Nasir2712:
can you remove WIP from the title if this is ready to be reviewed?

LeonidMarushevskyi:
I actually introduced constant for required label. Required label can be consider common business requirement which should be consistent across the application. Having hard coded '(required)' label across different files in application can be considered duplication.  Constant was introduced in attempt to localize this aspect of logic in one place. So if later PO will decide to change required label to * for example, it will be change in one place. I agree that there can be other better ways to have this logic in one place like using of styles but constant was the simplest way.

nasar-cambria:
@Nasir2712  this is not ready to be reviewed yet. I am still waiting on new version to be published from react-wood-duck to be able to see my changes in RFA.

nasar-cambria:
jenkins re-test this please

nasar-cambria:
Created a fresh PR to avoid test failures.

PR #404, opened 2018-03-30
CALS-5823 Facility tests - update with new facility data

sunilb0575:
- https://osi-cwds.atlassian.net/browse/CALS-5823

- Update facility search and profile rspec tests with new data.

- No

- rspec spec
- yarn spec-end2end-features

- No


Nasir2712:
@ratneshraval @sunilb0575  so I assume we will be having some kind of pipeline for deployment to staging env. So, don't you think if the data is different some of the test cases might fail which might halt deployment?

sunilb0575:
@Nasir2712 yes, I tried writing data independent test cases for few. But some features are difficult to test without exact data. 
It would have been better if all the environments have same data.

ratneshraval:
@Nasir2712 @sunilb0575 Staging may have different set of data but it may also have different set of tests. I think this approach is good till we figure out what is needed in staging.

Nasir2712:
@ratneshraval  I think we should comment out children and the complaint tests or any data dependent tests just because they will fail on staging. Can the test cases be switched based on environment? or is that even a good idea?

ratneshraval:
@Nasir2712 let's comment that out when it fails in staging, not right now. By then we'll figure out testing strategy for staging.

PR #397, opened 2018-03-23
WIP - FIT-1 Build image and run tests

ratneshraval:
#### Which User story does this relate to?
- 

#### Brief feature description
- 

#### Does this PR block/effect other stories?
- 

#### Do any commands needs to be run in order to run the project with this PR?
- 

#### Code coverage percent diff as a result of this PR?
-

PR #407, opened 2018-03-30
updated the node to "cals-slave"

apatil-osi:
- DevOps on-call related task

- Updated node definition to 'cals-slave'. All cals jobs were being run on master dev-jenkins box resulting in a cpu spike that caused the jenkins service to go down.

- No

- No

-


ratneshraval:
We don't need this anymore, as #397 will take care of this. I've disabled the job till then.

PR #411, opened 2018-04-02
CALS-5696 Removing (required) label from Employment Card

nasar-cambria:
- https://osi-cwds.atlassian.net/browse/CALS-5696

- Since the (required) label in address fields are coming from wood-duck its adding (required) label to every common address component in RFA 01A. We need to make sure this label is only added to fields which are required.

- no

- no

- NA


app/javascript/rfa_forms/rfa01a_edit_view/referencesCard.jsx
```diff
@@ -53,6 +53,7 @@ export default class ReferencesCard extends React.Component {
             index={this.props.index}
             stateTypes={this.props.stateTypes}
             addressTitle='Physical Address (required)'
```
anuroop15: why there are two required  texts? Both addressTitle and label attributes.

app/javascript/rfa_forms/rfa01b_edit_view/applicantDetailsCard.js
```diff
@@ -79,6 +79,7 @@ export default class ApplicantDetailsCard extends React.Component {
             stateTypes={this.props.stateTypes}
             addressTitle={'Residence Address' + RfaCommon.requiredIndicator}
             id='street_address'
+            label=' (required)'
```
anuroop15: can we use "RfaCommon.requiredIndicator" as we are already using in the file.

app/javascript/rfa_forms/rfa01a_edit_view/referencesCard.jsx
nasar-cambria: addressTitle is for Physical address and Label is being passed as props which is coming from wood-duck for ```City``` ```State``` and ```Zip Code```

app/javascript/rfa_forms/rfa01b_edit_view/applicantDetailsCard.js
nasar-cambria: As discussed in the earlier PR we would only want to use constant for lengthy labels/texts until we decide on implementing all the labels using that approach. So I would like to keep this as is for now.

anuroop15: Agreed for lengthy labels. But required case was different. We created a constant 'requiredIndicator' expecting a removal of required inputs in future.

nasar-cambria: I think creating constants leads to unnecessary render  ```https://www.kirupa.com/react/avoiding_unnecessary_renders.htm``` 

If we are thinking of an approach where we can make a change at one place and not on all the fields then having this inside of CSS would be better. Thoughts ?

ratneshraval: @nasar-cambria I don't think having a constant will cause an unnecessary render.
Although, I like your idea about keeping it in CSS so styles can be changed easily with text. But I don't think that work is needed yet, so current approach is fine with me.

nasar-cambria:
jenkins re-test this please

sunilb0575:
jenkins, retest this please

PR #414, opened 2018-04-03
Facility Profile Refactor

Nasir2712:
- n/a

- test

- no

- no

- n.a


config/routes.rb
```diff
@@ -3,6 +3,7 @@
 
   resources :facilities, only: [:index, :show] do
```
Nasir2712: the problem is page refresh, here if I add the post method, then I have to keep the show method as well page refresh.

PR #416, opened 2018-04-03
Submit button alias and mailing address CALS-5685

lauherkcambria:
- CALS-5685

- add alias and physical different than mailing address for submit validations.

- no

- no

- 85.39% Statements 76.29% Branches 81.52% Functions 85.82% Lines


app/javascript/rfa_forms/rfa01a_edit_view/addressCard.js
```diff
@@ -35,6 +53,10 @@ export default class AddressCard extends React.Component {
     this.props.setParentState('addresses', addresses.toJS())
   }
 
+  isPhysicalDifferentThanMailingAddress () {
+    return (this.props.physicalMailingSimilar == 'false' || this.props.physicalMailingSimilar == false)
```
nasar-cambria: Are we using this explicit check because we are expecting a JSON 'false' from the server ? if so, is there a way we can do ```JSON.parse()``` to convert such bool values in future or is that an overkill ? :)

lauherkcambria: it's a string 'false' onChange within the app and on save it comes back as a boo, which is why we need both. And although you're right, we definitely can use JSON.parse, the general consensus i've found while reading is that using JSON.parse is overkill for this kind of check. 👍

lauherkcambria:
jenkins, retest this please.

lauherkcambria:
@ratneshraval  I have created technical debt story https://osi-cwds.atlassian.net/browse/CALS-5879 to cover bringing the code coverage up as discussed.

app/javascript/rfa_forms/rfa01a_edit_view/completeNameField.jsx
```diff
@@ -1,72 +1,71 @@
...
-  validator
-}) => {
-  return (
-    <div>
-      {
-        prefixTypes && <div className='col-md-12'>
+export default class CompleteNameFields extends React.Component {
+  constructor (props) {
+    super(props)
+    this.props.validator && this.props.validator.addFieldValidation(this.props.idPrefix + 'first_name',
```
ratneshraval: Can you use one if and put both rule additions under that?

app/javascript/rfa_forms/rfa01a_edit_view/completeNameField.jsx
```diff
@@ -1,72 +1,71 @@
...
+      {rule: 'isRequired', message: 'Required'})
+    this.props.validator && this.props.validator.addFieldValidation(this.props.idPrefix + 'last_name',
+      {rule: 'isRequired', message: 'Required'})
+  }
+  componentWillUnmount () {
+    const rulesToRemove = [this.props.idPrefix + 'first_name', this.props.idPrefix + 'last_name']
+    this.props.validator && this.props.validator.removeValidations(rulesToRemove)
```
ratneshraval: Can use the validator check before the `const rulesToRemove` is created. I suggest wapping both statements in single validator check.

app/javascript/rfa_forms/rfa01a_edit_view/nameCard.js
```diff
@@ -26,21 +27,26 @@ export default class NameCard extends React.PureComponent {
...
   }
-
   addCard (event) {
     this.props.setParentState('other_names', addCardAsJS(this.props.nameFields.get('other_names'), Immutable.fromJS(blankNameFields)))
   }
 
   removeCard (indexValue) {
     this.props.setParentState('other_names', this.props.nameFields.get('other_names').delete(indexValue))
   }
+  isAliasPresent () {
```
ratneshraval: I don't see this function used anywhere.

app/javascript/rfa_forms/rfa01a_edit_view/addressCard.js
lauherkcambria: @ratneshraval, as we've discussed previously, the reason why #295 never got merged is because we are depending on the difference between strings and booleans as api sends us boolean and in the code we use string in the  handleClearOnConditionalChange methods.

app/javascript/rfa_forms/rfa01a_edit_view/completeNameField.jsx
lauherkcambria: 👍

app/javascript/rfa_forms/rfa01a_edit_view/completeNameField.jsx
lauherkcambria: 👍

app/javascript/rfa_forms/rfa01a_edit_view/nameCard.js
lauherkcambria: 👍

app/javascript/rfa_forms/rfa01a_edit_view/addressCard.js
Nasir2712: can you add one more ```=``` for a stricter check, code climate error

PR #417, opened 2018-04-03
Revert to working facility search

sunilb0575:


Nasir2712:
i showed my story to leon and he said its fine and put everything back to how it was.

PR #419, opened 2018-04-04
Add acceptance_tests

sunilb0575:
-Split acceptance tests from feature tests.
-Create a different directory for acceptance tests.


ratneshraval:
@sunilb0575 I will approve this for now but the folder names should be flipped. Acceptance tests should contain what we run in acceptance environment, not what we EXCLUDE.

PR #421, opened 2018-04-05
Remove Webpacker file loader

ratneshraval:
- None

- Remove image loader as we already append `/cals` for production assets and do not need to prepend it again for files path.


Nasir2712:
also, do you want me to create a new story to address the stylesheets issue for index level testing that was discussed previously?

ratneshraval:
Sure, add a research story on why it isn't loading properly

ratneshraval:
retest this please

PR #426, opened 2018-04-09
Freeform relationship to applicant CALS-5731

lauherkcambria:
- https://osi-cwds.atlassian.net/browse/CALS-5731

- replace current dropdown for relationship types to freeforms in minor children and other adults cards. 

(**Note:** As this story is merely changing the type of these 2 fields from dropdowns to freeform text fields, for testing purposes only the previously created tests on those fields were fixed to test the change. Further tests on the validation object will occur as apart of technical debt story: https://osi-cwds.atlassian.net/browse/CALS-5879)

- no

- yarn rebuild-assets-dev

- 85.29% Statements
74.87% Branches 
81.72% Functions 
85.71% Lines


lauherkcambria:
jenkins, retest this please.

app/controllers/rfa/a01_controller.rb
```diff
@@ -114,7 +114,7 @@ def minor_children_params
       minor['relationship_to_applicants'][0] = set_relationship_to_applicants(minor['relationship_to_applicants'][0], @application_response[:applicants])
       ActionController::Parameters.new(minor.to_h).permit(
         :id, :to_delete, :child_financially_supported, :date_of_birth, :child_adopted, gender: %i[id value],
-        relationship_to_applicants: [:applicant_id, relationship_to_applicant: %i[id value]]
+        relationship_to_applicants: [:applicant_id, :relationship_to_applicant_freeform, relationship_to_applicant: %i[id value]]
```
Nasir2712: do we still need ```relationship_to_applicant: %i[id value]``` since we moved to just input field?
same question for below method as well.

app/javascript/components/common/OtherAdultsCardField.js
```diff
@@ -61,13 +61,11 @@ export class OtherAdultsCardField extends React.Component {
...
-          id={this.props.idPrefix + 'relationshipType'}
-          selectClassName='reusable-select'
-          optionList={this.props.relationship_types}
-          label={isRelationshipToApplicantObject ? 'Relationship Type (required)' : 'Relationship Type'}
-          value={getDictionaryId(adult.relationship_to_applicants[0].relationship_to_applicant)}
-          onChange={(event) => this.props.handleRelationshipTypeToApplicant(this.props.index, dictionaryNilSelect(event.target.options), 'relationship_to_applicant')} />
+        <InputComponent gridClassName='col-md-4'
+          id={this.props.idPrefix + 'relationship_to_applicant_freeform'}
+          value={adult.relationship_to_applicants && adult.relationship_to_applicants[0] ? adult.relationship_to_applicants[0].relationship_to_applicant_freeform : ''}
+          label={'Relationship to Applicant'}
```
Nasir2712: please remove braces around strings.

app/javascript/helpers/validator.js
```diff
@@ -85,6 +87,25 @@ export default class Validator {
...
+    return true
+  }
+
+  isRequiredIfNumber (opt) {
+    if (opt.condition()) {
+      if (opt.value !== '') {
+        if (typeof (opt.value) === 'string') {
+          return !_.isEmpty(opt.value.trim())
+        } else {
+          return (!isNaN(+opt.value))
```
Nasir2712: can you let us know what is ```+opt```?

app/javascript/rfa_forms/rfa01a_edit_view/OtherAdultsCardsGroup.js
```diff
@@ -7,10 +7,14 @@ import {addCardAsJS, removeCardWithId, handleRelationshipTypeToApplicant, getFoc
 export const otherAdultsDefaults = Object.freeze({
   relationship_to_applicants: [
     {
+      relationship_to_applicant_freeform: '',
```
Nasir2712: can you move these defaults to ```defaultFields.js```?

app/javascript/rfa_forms/rfa01a_edit_view/OtherAdultsCardsGroup.js
```diff
@@ -7,10 +7,14 @@ import {addCardAsJS, removeCardWithId, handleRelationshipTypeToApplicant, getFoc
...
       applicant_id: '',
       relationship_to_applicant: null
```
Nasir2712: also, do we still need ```relationship_to_applicant```?

app/javascript/rfa_forms/rfa01a_edit_view/addressCard.js
```diff
@@ -54,7 +54,7 @@ export default class AddressCard extends React.Component {
   }
 
   isPhysicalDifferentThanMailingAddress () {
-    return (this.props.physicalMailingSimilar == 'false' || this.props.physicalMailingSimilar == false)
+    return (this.props.physicalMailingSimilar === 'false' || this.props.physicalMailingSimilar === false)
```
Nasir2712: 👍

app/javascript/rfa_forms/rfa01a_edit_view/minorCardField.js
```diff
@@ -63,13 +64,11 @@ export class MinorCardField extends React.Component {
...
-        <DropDownField gridClassName='col-md-4'
-          id='relationship_to_applicant'
-          selectClassName='reusable-select'
-          optionList={this.props.relationshipToApplicantTypes}
-          value={getDictionaryId(minor.relationship_to_applicants[0].relationship_to_applicant)}
-          label={isRelationShipToApplicantObject ? 'Relationship Type (required)' : 'Relationship Type'}
-          onChange={(event) => this.props.handleRelationshipTypeToApplicant(this.props.index, dictionaryNilSelect(event.target.options), 'relationship_to_applicant')} />
+        <InputComponent gridClassName='col-md-4'
+          id={this.props.idPrefix + 'relationship_to_applicant_freeform'}
+          value={minor.relationship_to_applicants && minor.relationship_to_applicants[0] ? minor.relationship_to_applicants[0].relationship_to_applicant_freeform : ''}
```
Nasir2712: can you wrap this in a separate method that can be used for both minorCardField and other adults?

app/javascript/rfa_forms/rfa01a_edit_view/minorCardsGroup.jsx
```diff
@@ -8,6 +8,7 @@ import {addCardAsJS, removeCardWithId, handleRelationshipTypeToApplicant, getFoc
 export const minorDefaults = Object.freeze({
   relationship_to_applicants: [
     {
+      relationship_to_applicant_freeform: '',
       applicant_id: '',
       relationship_to_applicant: null
```
Nasir2712: if you can move these defaults to ```defaultFields.js```, that would be great.

app/javascript/components/common/OtherAdultsCardField.js
```diff
@@ -61,13 +61,11 @@ export class OtherAdultsCardField extends React.Component {
...
-        <DropDownField gridClassName='col-md-4'
-          id={this.props.idPrefix + 'relationshipType'}
-          selectClassName='reusable-select'
-          optionList={this.props.relationship_types}
-          label={isRelationshipToApplicantObject ? 'Relationship Type (required)' : 'Relationship Type'}
-          value={getDictionaryId(adult.relationship_to_applicants[0].relationship_to_applicant)}
-          onChange={(event) => this.props.handleRelationshipTypeToApplicant(this.props.index, dictionaryNilSelect(event.target.options), 'relationship_to_applicant')} />
+        <InputComponent gridClassName='col-md-4'
+          id={this.props.idPrefix + 'relationship_to_applicant_freeform'}
+          value={adult.relationship_to_applicants && adult.relationship_to_applicants[0] ? adult.relationship_to_applicants[0].relationship_to_applicant_freeform : ''}
```
anuroop15: can we avoid lengthy array check by using 'checkArrayObjectPresence' helper method

app/javascript/components/common/OtherAdultsCardField.js
```diff
@@ -30,7 +30,7 @@ export class OtherAdultsCardField extends React.Component {
         condition: () => this.isRelationshipToApplicantObject()})
     this.props.validator.addFieldValidation(
       this.ApplicantIdID,
-      {rule: 'isRequiredIf',
+      {rule: 'isRequiredIfNumber',
```
ratneshraval: Looks like you're trying to check that, if condition meets the value should be number. 
I think it is more readable `isRequiredNumberIf`
Also you can probably stop at `this.props.idPrefix + 'relationship_to_applicants[0]` and maybe use `isRequiredObjectIf` if the need is to check if dropdown has value selected or not.

app/javascript/components/common/OtherAdultsCardField.js
ratneshraval: Can you also add defaultProp for that? Maybe it will help simplify the check.

app/javascript/helpers/validator.js
ratneshraval: Is there a case when incoming value is of type `string`? What's the reason to check for string here?

nasar-cambria: Quick question -  if we are expecting some value with string then would it be a good idea to use ```Number.isNaN() method``` that way we are checking both value and type?

app/controllers/rfa/a01_controller.rb
lauherkcambria: yes, This freeform solution is temporary so I don't want to remove that as it may come back in the future.

app/javascript/components/common/OtherAdultsCardField.js
lauherkcambria: 👍

app/javascript/helpers/validator.js
lauherkcambria: 1.@Nasir2712  +opt is a unary plus operator in javascript. According to the javascript docs,  Unary plus is the fastest and preferred way of converting something into a number, because it does not perform any other operations on the number https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Arithmetic_Operators#Unary_plus_()

2. Yes, There is a case when it is a string, on first application load the applicants have yet to be created until a first save. So, when the applicant is selected from the to whom dropdown it will be a string. 

3. we can check, that may also be another way of doing it as well

app/javascript/rfa_forms/rfa01a_edit_view/OtherAdultsCardsGroup.js
lauherkcambria: 👍

app/javascript/rfa_forms/rfa01a_edit_view/OtherAdultsCardsGroup.js
lauherkcambria: yes. As This freeform solution may be temporary, we don't want to remove it as it may come back in the future. Besides, it is still apart of api structure.

app/javascript/rfa_forms/rfa01a_edit_view/minorCardsGroup.jsx
lauherkcambria: 👍

app/javascript/rfa_forms/rfa01a_edit_view/minorCardField.js
lauherkcambria: i'm not quite sure I follow the purpose of that, it seems like unnecessary abstraction.

app/javascript/components/common/OtherAdultsCardField.js
lauherkcambria: we don't need to add default props here as we are adding the default props to the group card.

app/javascript/components/common/OtherAdultsCardField.js
lauherkcambria: I'm not q;uite sure I follow this comment-we needed to add a validator method for the case of numbers as _.isEmpty() isn't robust enough of a check. So is the jist of this comment that you want this method to be renamed to 'isRequiredNumberIf' ?

lauherkcambria:
@ratneshraval for last two comments, these are repeats and have been addressed throughout this pr.

ratneshraval:
@lauherkcambria can you update the story number in PR title?

app/javascript/rfa_forms/rfa01a_edit_view/minorCardField.js
Nasir2712: I don't think it's an unnecessary abstraction, because we are using the same kind of checks in three places now.
1) in ```minorCardField.js - minor.relationship_to_applicants && minor.relationship_to_applicants[0] ? minor.relationship_to_applicants[0].relationship_to_applicant_freeform : '' ```
2) in ```OtherAdultsCardField.js- adult.relationship_to_applicants && adult.relationship_to_applicants[0] ? adult.relationship_to_applicants[0].relationship_to_applicant_freeform : '' ```
3) also in ```desiredChildRelationships.js```

lauherkcambria: so what you mean is not an abstraction of the input field, but a function to make these values cleaner.

app/javascript/helpers/validator.js
ratneshraval: If you use loadash `_.isNumber` you can get rid of all the type and string checks.

lauherkcambria: did you try using  `_.isNumber` in place of this code? I can't seem to get this method to work.

app/javascript/components/common/OtherAdultsCardField.js
lauherkcambria: ok, i'll rename `isRequiredIfNumber` to be `isRequiredNumberIf`

app/javascript/helpers/validator.js
ratneshraval: You're right, it doesn't work because we have numbers coming in string format. What you have is fine till we can find a different solution.

app/javascript/rfa_forms/rfa01a_edit_view/minorCardField.js
```diff
@@ -31,25 +32,25 @@ export class MinorCardField extends React.Component {
...
         message: 'required',
         condition: () => this.isRelationShipToApplicantObject()})
     this.props.validator.addFieldValidation(this.childFinanciallySupportedID,
-      {rule: 'isRequiredIf',
+      {rule: 'isRequiredIfBoolean',
         message: 'required',
         condition: () => this.isRelationShipToApplicantObject()})
     this.props.validator.addFieldValidation(this.childAdoptedID,
-      {rule: 'isRequiredIf',
+      {rule: 'isRequiredIfBoolean',
```
ratneshraval: Can you also rename these to be `isRequiredBooleanIf`?

lauherkcambria: renaming 'isRequiredIfBoolean' to 'isRequiredBooleanIf'.

lauherkcambria:
this was merged and deleted before I finished addressing nesh's comments.

PR #434, opened 2018-04-14
CALS-5862: CALS Implementation - Add default logout success page

anuroop15:
- https://osi-cwds.atlassian.net/browse/CALS-5862

- updated logout variable with latest version pf perry URL

- NA

- NA

- increased header/index.js coverage


ratneshraval:
Check #387

lauherkcambria:
jenkins, retest this please.

test/javascript/header/index_tests.js
```diff
@@ -1,13 +1,29 @@
...
     expect(header.props().logoutUrl).toEqual('/logout')
     header.setProps({ logoutUrl: '/abcd' })
     expect(header.props().logoutUrl).toEqual('/abcd')
   })
+  it('allow user logout', () => {
+    let avatar = header.find('.profile-avatar a')
+    avatar.simulate('click')
+    let logOutLink = header.find('ul.c_dropdown a')
+    logOutLink.simulate('click')
+    expect(logoutCallbackSpy()).toBe('/logout')
```
ratneshraval: This is a pointless expect statement since you're setting the return value in line 11.

test/javascript/header/index_tests.js
```diff
@@ -1,13 +1,29 @@
...
+    expect(header.instance().redirectUrl).toHaveBeenCalled()
```
ratneshraval: Still not testing `toHaveBeenCalledWith`.

Nasir2712:
can you remove WIP from the title, and also update the title of the PR to match story title.

test/javascript/header/index_tests.js
anuroop15: doesn't have to be a rude comment here. If you don't get the point I can explain it to you.
The the reason why I'm expecting a return value is that I have mocked the logoutCallback method to return a string. 
The reason I did this is the page refreshes on logout button click and chrome will reload the page making the karma runner to be disconnected, so I have mocked the existing method.

test/javascript/header/index_tests.js
anuroop15: why does it need a toHaveBeenCalledWith if it's testing method. Any reason your expecting arguments here?

config/initializers/1_ENV_VARS.rb
```diff
@@ -1,5 +1,6 @@
 CALS_API_BASE_URL = ''
 CALS_API_URL = ENV['CALS_API_URL']
+CALS_LOGOUT_URL=ENV['CALS_LOGOUT_URL']
```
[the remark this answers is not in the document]
anuroop15: A new url was added for Perry log out url. I have added that url in env file as mentioned in the PR description.

app/javascript/header/index.js
```diff
@@ -7,8 +7,13 @@ export default class HeaderComponent extends React.Component {
     this.logoutCallback = this.logoutCallback.bind(this)
   }
 
+  redirectUrl (path) {
+    window.location.href = path
+    return window.location.href
```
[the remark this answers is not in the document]
anuroop15: instead of assigning directly a href to window object, I added path to window object through dependency injection

config/initializers/1_ENV_VARS.rb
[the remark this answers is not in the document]
ratneshraval: This doesn't need to be an env variable. We have Perry base url. You need to append to that, just how we are doing with login url.

test/javascript/header/index_tests.js
ratneshraval: @anuroop15 it is a pointless expect statement that doesn't do anything. you can change the code of the method and your test will still pass. You can change it to `/xyz` in line 11 and line 26 and it will still pass.
Can you explain what is the point of such `expect` statement?

anuroop15: this test is not expecting a return value of logoutCallback. I'm testing whether the logout method was called on 'Logout' anchor tag click or not. There is nothing to do with method's return value.

config/initializers/1_ENV_VARS.rb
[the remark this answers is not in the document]
anuroop15: Will update with Perry base url.

ratneshraval:
@anuroop15 as discussed, remove the `callFake`, return value from test and update code file to remove return value as well

app/controllers/cals_base_controller.rb
```diff
@@ -7,7 +7,7 @@ class CalsBaseController < ApplicationController
 
   def logout
     delete_user_from_session
-    default_callback_url = ''
+    default_callback_url = request.url
```
ratneshraval: This doesn't seem to be doing anything. What's the purpose of this?

anuroop15: API is expecting a callback variable. so I'm passing request url instead of string. Do you want me to change it back it string??

anuroop15:
sure. Agreed. will revert back to empty string.

anuroop15:
jenkins retest this please

PR #437, opened 2018-04-17
Add faker to rfa 01a

sunilb0575:
- Na. 

-Test case was flaky, was failing once out of 3 times.
-Remove blocker for pipeline deployment into integration. 

-No

-TEST_END_TO_END=true SELENIUM_BROWSER=CHROME bundle exec rspec spec/acceptance/rfa_forms/rfa_01a_application_spec.rb:63

-No


ratneshraval:
To me, this test does not belong in feature test. We should verify this in karma, if needed.

PR #443, opened 2018-04-23
CALS-5832-blank page listview

sunilb0575:
-https://osi-cwds.atlassian.net/browse/CALS-5832

-Component was breaking when county value is null, Check condition added.

-No

-No

-Functions - 82


test/javascript/search/searchList_tests.js
```diff
@@ -1,89 +1,124 @@
...
+
+  it('Verify facility address is N/A', () => {
+    expect(searchListComp.find('#address_row1').props().children).toBe('N/A')
+  })
+
+  it('Verify facility phone number rendered', () => {
+    expect(searchListComp.find('#phone_row0').props().children).toBe('(831) 763-2735')
+  })
+
+  it('Verify facility phone number is  N/A', () => {
```
Nasir2712: can you write a test case for the county to check if it's null and when it has value?

app/javascript/search/searchList.js
```diff
@@ -30,12 +30,12 @@ export const SearchList = ({
             <tr key={index}>
               <td><a href={urlPrefixHelper('/facilities/' + result.id)}>{result.name}</a></td>
               <td>{result.license_number}</td>
-              <td>{checkForNA(result.type) + ' / ' + checkValueForNull(result.facility_source)}</td>
+              <td id={'type_row' + index}>{checkForNA(result.type) + ' / ' + checkValueForNull(result.facility_source)}</td>
```
ratneshraval: Lets switch this identifier to be more readable. Something like `facilities0.type`

PR #446, opened 2018-04-25
CALS-5956 User with no authorized privileges can access facility search using URL

Nasir2712:
- https://osi-cwds.atlassian.net/browse/CALS-5956

- Cals User not having enough privileges to access facility search and profile can still access the application using URL.

- n/a

- n/a

- n/a


anuroop15:
Can this have an acceptance test, to check the direct url.

Nasir2712:
@ratneshraval comments addressed.

PR #448, opened 2018-04-27
CALS-5734: Applicant Full Name Shown As Label on RFA 01B

anuroop15:
- https://osi-cwds.atlassian.net/browse/CALS-5734

- the applicant name to appear as a label on the RFA 01B form so that I know which B form

- NA

- NA

- NA


app/javascript/components/common/pageHeader.js
```diff
@@ -5,6 +5,7 @@ import Affix from 'react-overlays/lib/AutoAffix'
 
 const PageHeader = ({
   headerLabel,
+  applicantName,
```
ratneshraval: let's make this generic by naming it `headerLine2` or `headerLabelRow2`.

app/javascript/components/common/pageTemplate.js
```diff
@@ -66,6 +69,7 @@ PageTemplate.propTypes = {
 
 PageTemplate.defaultProps = {
   headerLabel: 'To implement',
+  applicantName: undefined,
```
Nasir2712: can this just be an empty string? since the prop-type for applicant name is a string, this can't be undefined.

app/javascript/rfa_forms/rfa01b_edit_view/index.js
```diff
@@ -173,13 +175,18 @@ export default class Rfa01bEditView extends React.Component {
     return this.state.activeNavLinkId === id
   }
 
+  getApplicantFullName (applicants, otherAdults) {
```
Nasir2712: I don't see other adults being used, did you miss them?

Nasir2712: Also, a suggestion, see if you can move that this method out to a common helper, and call the method directly when rendering the props to make index cleaner.

app/javascript/components/common/pageTemplate.js
anuroop15: I believe there is no such rule that states a prop type specified as string can't be defaulted as undefined. 
As undefined is a primitive value used when a variable has not been assigned a value, we can default it to undefined if we don't specify is as required in propTypes or so.

app/javascript/rfa_forms/rfa01b_edit_view/index.js
anuroop15: I have removed other adults argument.
I appreciate the suggestion, but this method can't be reused in other cards as it reads applicant_first_name property from argument. but most of the card in RFA have first_name property.

app/javascript/components/common/pageTemplate.js
ratneshraval: defaultProps trigger when a prop coming in is undefined. You don't have to make anything undefined in default props.

app/javascript/components/common/pageTemplate.js
```diff
@@ -66,6 +69,7 @@ PageTemplate.propTypes = {
 
 PageTemplate.defaultProps = {
   headerLabel: 'To implement',
+  pageSubHeader: undefined,
```
ratneshraval: Remove this `pageSubHeader: undefined`

anuroop15: Any reason for this to be removed? I made it undefined intentionally so that it will not add a extra div while loading the DOM. 
if it's removed, name will have empty string, which will add empty div to the page header.

Nasir2712: I think if you wrap the div inside the conditional statement you could still stop rendering the extra div.

anuroop15: as per the business rule, we don't need a conditional check here. So removing the default props logic here.

anuroop15:
jenkins retest this please

app/javascript/rfa_forms/rfa01b_edit_view/index.js
```diff
@@ -11,7 +11,7 @@ import CrimeBackgroundAgainstCohabitant from './crimeBackgroundAgainstCohabitant
 import PrivacyStatement from './privacyStatement'
 import RfaSideBar from 'rfa_forms/rfa_sidebar/index.js'
 import {CountyUseOnlyCard} from 'components/rfa_forms/countyUseOnlyCard.js'
-import {getDictionaryId, dictionaryNilSelect, checkArrayObjectPresence} from 'helpers/commonHelper.jsx'
+import {getDictionaryId, dictionaryNilSelect, checkArrayObjectPresence, getNameById} from 'helpers/commonHelper.jsx'
```
sunilb0575: If ```getNameById``` is not used anywhere you can remove it.

sunilb0575:
Everything looks good to me.

PR #461, opened 2018-05-07
SEAR-75 Sort search results based on facility name

sunilb0575:
-https://osi-cwds.atlassian.net/browse/SEAR-75

-Sort facility search results by facility name. 

-No

-No

-No


Nasir2712:
if you can update the ```query_builder_spec.rb``` file to include sort by name tests, that would be helpful, If not you can just update the current tests with name instead of ``` _score```.

sunilb0575:
@Nasir2712 Changed two of the tests to test sort by 'name', left one as is, to test sort by '_score'.

Nasir2712:
any chance you can rebase?

sunilb0575:
@Nasir2712 already did, want to keep that second commit, as it's addressing change in test case.

Nasir2712:
ya, it is a part of this story so it should be rebased.

sunilb0575:
ok

app/javascript/sagas/fetchSearchResultsSaga.js
```diff
@@ -7,7 +7,7 @@ import {SEARCH_RESULTS_FETCH, fetchSuccess, fetchFailure} from 'actions/searchAc
 // worker saga: makes the api call when watcher saga sees the action
 export function * fetchSearchResults (action) {
   try {
-    const url = '/facilities/search' + '?from=' + action.payload.urlParams.fromValue + '&size=' + action.payload.urlParams.sizeValue
+    const url = '/facilities/search' + '?from=' + action.payload.urlParams.fromValue + '&size=' + action.payload.urlParams.sizeValue + '&sort=' + action.payload.urlParams.sort + '&order=' + action.payload.urlParams.order
```
ratneshraval: Maybe declare `const = action.payload` so you don't have to keep chaining everywhere in the method.

PR #463, opened 2018-05-07
Yarn lint to raise errors

ratneshraval:
- n/a

- `yarn lint` will now raise errors instead of showing errors but returning success


ratneshraval:
@Nasir2712 it is checking both `app` and `test`

PR #465, opened 2018-05-09
Sear 84 display client number

AlexanderSerbin:
https://osi-cwds.atlassian.net/browse/SEAR-84

Facility Profile should display 19 digit client ID number that is provided for the user in CWS/CMS

No

No

0%


Nasir2712:
can you change the title of the PR, so that it matches the name of your story?

app/javascript/facility/facilityChildren.jsx
```diff
@@ -11,7 +11,7 @@ const FacilityChildren = ({
       <table id='facility-children-table' className='table'>
         <thead>
           <tr>
-            <th> ID </th>
+            <th> CLIENT NUMBER </th>
```
Nasir2712: Just talked to PO about this. The label should be CLIENT ID not CLIENT NUMBER.

AlexanderSerbin:
Changed PR title to SEAR-84 Facility Profile should display 19 digit client ID number that is provided for the user in CWS/CMS

PR #469, opened 2018-05-11
SEAR-44 Switching elasticsearch query to match and default sort based on score

ratneshraval:
- https://osi-cwds.atlassian.net/browse/SEAR-44

- Elasticsearch query builder uses "match" queries instead of "match_phrase"
- Sorting based on score


Nasir2712:
rebasing needed to merge.

PR #477, opened 2018-05-15
SEAR-123 Add bottom pagination

sunilb0575:
- https://osi-cwds.atlassian.net/browse/SEAR-123

- Add bottom pagination on facility search

-No

-No

-Branches 79


app/javascript/search/search.js
```diff
@@ -78,6 +78,18 @@ class Search extends React.Component {
           {!this.props.isToggled && <SearchList searchResults={this.props.searchResults} />}
           {(!searchResponseHasValues && !initialLoad) && <SearchNotFound errors={this.props.errors.issue_details} errorMessage={this.props.errorMessage} />}
         </div>
+        <div className='search_details col-xs-12 col-sm-12 col-md-12 col-lg-12'>
+          {searchResponseHasValues &&
+          <Pagination
+            paginationClassName={'bottom_pagination'}
```
Nasir2712: remove braces around string

app/javascript/search/search.js
```diff
@@ -78,6 +78,18 @@ class Search extends React.Component {
           {!this.props.isToggled && <SearchList searchResults={this.props.searchResults} />}
           {(!searchResponseHasValues && !initialLoad) && <SearchNotFound errors={this.props.errors.issue_details} errorMessage={this.props.errorMessage} />}
         </div>
+        <div className='search_details col-xs-12 col-sm-12 col-md-12 col-lg-12'>
```
Nasir2712: render this div only when you have results or else use this div inside the Pagination Component.

app/javascript/search/searchDetails.js
```diff
@@ -2,56 +2,34 @@ import React from 'react'
...
-      </button>
-    </div>
-    <div className='toggle_result col-xs-12 col-sm-3 col-md-3 col-lg-3'>
-      <div className='pull-right'>
-        <div id='toggle_button' onClick={handleToggle} className={(toggeledResult ? 'line_off-icon' : 'line_on-icon') + ' ' + 'navbar-brand'} alt={'list'} />
-        <div onClick={handleToggle} className={(toggeledResult ? 'grid_on-icon' : 'grid_off-icon') + ' ' + 'navbar-brand'} alt={'grid'} />
-      </div>
-    </div>
+      <Pagination
+        paginationClassName={'top_pagination'}
```
Nasir2712: remove braces around string

test/javascript/search/search_tests.js
```diff
@@ -108,7 +108,41 @@ describe('Verify Search component', function () {
   it('verify clicking search button calls handleOnSubmit method', () => {
     let searchFacility = searchComp.find('#search')
     searchFacility.simulate('submit')
-    expect(searchApiCallSpy).toHaveBeenCalledWith({'county.id': '19', 'type.id': undefined, 'id': undefined, 'name': undefined, 'addresses.address.street_address': undefined}, {fromValue: 0, sizeValue: 20, sort: 'name.for_sort', order: 'asc'})
+    expect(searchApiCallSpy).toHaveBeenCalledWith({'county.id': '19', 'type.id': undefined, 'id': undefined, 'name': undefined, 'addresses.address.street_address': undefined}, {fromValue: 0, sizeValue: 5, sort: 'name.for_sort', order: 'asc'})
+  })
+
+  it('verify clicking next button ', () => {
```
Nasir2712: Instead of having these test cases used in the search component. I think it's better to test the Pagination component which would be more unit testing.

app/javascript/search/pagination.js
```diff
@@ -0,0 +1,64 @@
...
+      id={'next_button_' + paginationClassName}
+      disabled={getFromValue(sizeValue, pageNumber) + sizeValue >= totalNoOfFacilities}
+      onClick={() => { handlePageNumberChange(pageNumber + 1); searchApiCall(getFromValue(sizeValue, pageNumber + 1), sizeValue); window.scrollTo(0, 0) }}
+      className='next btn btn-default'>
+      <p>&#8250;</p>
+    </button>
+  </span>
+)
+
+Pagination.propTypes = {
```
Nasir2712: Please check these prop-types and default props they don't match the props you are sending.

sunilb0575:
jenkins, retest this please

Nasir2712:
jenkins retest this please

test/javascript/search/search_tests.js
sunilb0575: I thought testing child component after mounting it's parent component is better, anyways @Nasir2712  added pagination_tests too.

Nasir2712: So, the entire purpose of this refactor is to understand the way we are testing. So you have pagination functionality that you need to unit test. Search component purpose is to only check if the components are being rendered.  Now what you did is actually replicate test cases in two places which we should avoid. 

Remove the test cases from search component and only check if the pagination component is being rendered or not after the search results. Have all the test cases about previous page, next page in pagination component.

test/javascript/search/pagination_tests.js
```diff
@@ -0,0 +1,85 @@
+import React from 'react'
```
Nasir2712: I see the test cases are being replicated in both the describe blocks. 
What I would suggest is to have two separate describe blocks. One ```describe``` block should check pagination class names as props in two separate ```it``` statements and check if the component is being rendered or not. The other ```describe``` block should only check the functionality such as previous and next page.

sunilb0575: Removed second describe block and the test cases for bottom pagination and also added an it-block to verify render.

test/javascript/search/search_tests.js
sunilb0575: The reason to keep those test cases in the parent component is to test the method searchApiCall() which is being called from the child component. As we tested search button from searchInput component.

PR #486, opened 2018-05-18
AARA-151:  relationship between applicant only appearing  for 2 applicans

lauherkcambria:
-https://osi-cwds.atlassian.net/browse/AARA-151

- 
address bug where the relationship between applicant only appearing exactly when 2 applicants are present;we need it to show for any number of applicants >= 2. add supporting test cases.

-no

-no

- karma: 86.99% Statements 
                79.07% Branches
                84.32% Functions 
                87.21% Lines 
-rspec: All Files (93.0% covered at 171.26 hits/line)


spec/acceptance/rfa_forms/rfa_01a_application_spec.rb
```diff
@@ -286,6 +299,46 @@
...
+    select 'Other', from: 'relationship_type'
+    fill_in('other_relationship', with: 'test', match: :prefer_exact)
+    click_button('Save Progress')
+    expect(find_field('relationship_type').value).to eq '5'
+    expect(find_field('other_relationship').value).to eq 'test'
+    visit page.driver.current_url
+    select 'Cohabitants', from: 'relationship_type'
+    click_button('Save Progress')
+  end
+
```
nasar-cambria: @lauherkcambria instead of this can we write karma test cases I think that would be more helpful to catch any bugs if something breaks.

lauherkcambria: @nasar-cambria let's keep this spec test for functionality testing , and in addition as discussed i'll add an extra karma test. 👍 😄

nasar-cambria: cool 👍

PR #487, opened 2018-05-18
SEAR-129 Spike: Show/Hide bottom pagination based on number of search results 

Nasir2712:
-https://osi-cwds.atlassian.net/browse/SEAR-129

- Added conditional pagination dropdown based on the scrollbar

- no

- no

-n/a


app/javascript/search/common/commonUtils.js
```diff
@@ -86,6 +86,10 @@ export const checkNullOrEmptyValue = (value) => {
   return (value === null || value === '') ? 'N/A' : value
 }
 
+export const checkIfScrollBarVisible = (object) => {
```
ratneshraval: This isn't search specific. Can we move this function to more general file?

Nasir2712: Haven't used it anywhere else but except here. So, that's why had it here in search specific common utils.

Nasir2712:
jenkins, retest this please

ratneshraval:
I like the functionality and it works pretty well. Although I suggest we get this reviewed by PO.

My question is:
1. Should it decide on pagination when component updates
OR
2. Should it show/hide based on window resize, without component update.

Nasir2712:
The pagination is decided when the component updates. The pagination is dependent on Search Results. It cannot show/hide on window resize without a component update because whenever search results update it should check the presence of the scrollbar.

PR #493, opened 2018-05-22
AARA-264 Restrict input characters exceeding maximum length

nasar-cambria:
- https://osi-cwds.atlassian.net/browse/AARA-264

- RFA 01A, B, and C forms - User should not be allowed to input characters exceeding maximum length

- No

- No

-  NA


anuroop15:
Please merge after build go green.

PR #495, opened 2018-05-23
AARA-156: Values entered for married and domestic partnership relation-type persists for other type of relation-types 

lauherkcambria:
-https://osi-cwds.atlassian.net/browse/AARA-156

-handle data being persisted when the relationship type changes in relationship between applicants card

-no

- no

-87.18% Statements 
79.6% Branches 
84.46% Functions 
87.42% Lines


lauherkcambria:
jenkins, retest this please.

lauherkcambria:
jenkins, retest this please.

spec/acceptance/rfa_forms/rfa_01a_application_spec.rb
```diff
@@ -299,39 +299,24 @@
...
+  click_button('Add Another Applicant +')
+  fill_in('applicants[1].first_name', with: Faker::Name.first_name, match: :prefer_exact)
+  fill_in('applicants[1].last_name', with: Faker::Name.last_name, match: :prefer_exact)
+  select 'Married', from: 'relationship_type'
+  expect(find_field('relationship_type').value).to eq '1'
+  fill_in('date_of_relationship', with: '12/12/1211', match: :prefer_exact)
+  expect(find_field('date_of_relationship').value).to eq '12/12/1211'
+  select 'Domestic Partnership', from: 'relationship_type'
+  expect(find_field('date_of_relationship').value).to eq ''
+end
```
nasar-cambria: Looks Good. Would it make sense when a user selects ```relationship_type``` as ```Domestic Partnership``` the expect value should be ```.not_to have_content``` instead of ```.value).to eq ''```

Also, we need to have ```fill_in``` for two more fields ex: ```City``` and ```State```

app/javascript/rfa_forms/rfa01a_edit_view/relationshipBetweenApplicantsCard.jsx
```diff
@@ -53,6 +53,6 @@ export default class RelationshipBetweenApplicantsCard extends React.Component {
 }
 
 RelationshipBetweenApplicantsCard.defaultProps = {
-  relationshipBetweenApplicants: blankValues,
+  relationshipBetweenApplicants: blankRelationshipBetweenApplicantsValues,
```
nasar-cambria: 👍

spec/acceptance/rfa_forms/rfa_01a_application_spec.rb
lauherkcambria: good call on adding the other fill in fields -i believe `.not_to have_content` instead of `.value).to eq ''` are equivalent, but definitely easier to read and will make that change. 👍

lauherkcambria: turns out that `.not_to have_content` won't work because react expects a `''` instead of a `null` when setting the variable, so the tests actually need to expect `''`. good thought though, still added the other variables. cheers! 🍻

lauherkcambria:
jenkins, retest this please.

lauherkcambria:
jenkins, retest this please.

app/javascript/rfa_forms/rfa01a_edit_view/relationshipBetweenApplicantsCard.jsx
```diff
@@ -18,6 +12,12 @@ export default class RelationshipBetweenApplicantsCard extends React.Component {
 
   setState (key, value) {
     let newData = Immutable.fromJS(this.props.relationshipBetweenApplicants)
+    if (key === 'relationship_type') {
+      newData = newData.set('other_relationship', '')
```
anuroop15: you can use 
if (key === 'relationship_type') {
      newData = newData.set(blankRelationshipBetweenApplicantsValues)
    }

this will reduce the number of line of code.

lauherkcambria: good idea 👍

lauherkcambria:
jenkins, retest this please.

PR #506, opened 2018-05-30
Restrict number of digits for pagination input

sunilb0575:
-Restrict pagination input box to accepts maximum of 5 digits.


app/javascript/search/pagination.js
```diff
@@ -35,7 +35,7 @@ const Pagination = ({
       className='page_number'
       value={pageNumber}
       type='number'
```
Nasir2712: why can't you just simply do maxLength='5'?

sunilb0575: tried not working.

Nasir2712: how about just this``` if (event.target.value.length > 5) return false; ``` in the onchange method? also max length is only for type='text'.

ratneshraval: Let's try cleave.js and see if that helps.

PR #510, opened 2018-05-31
WIP- SEAR-194-Facility profile - load behavior change and loading indicator

Nasir2712:
- https://osi-cwds.atlassian.net/browse/SEAR-194

- clean up + loading indicator

- no

- no

-


Nasir2712:
My build is failing because my rspec is below its coverage. I commented out some code for children and complaints, once they are fixed I'll merge them in my branch and the merge this.

config/routes.rb
```diff
@@ -3,7 +3,8 @@
 
   resources :facilities do
     collection { post :search }
-    collection { post :facility }
+    #collection { post :facility }
+    collection { post :facility_profile}
```
ratneshraval: This should look like `facilities/:id/profile`,so add `get 'profile'` route.

app/javascript/containers/facilityAddressContainer.js
```diff
@@ -0,0 +1,11 @@
+import {connect} from 'react-redux'
+import FacilityAddress from 'facility/facilityAddress'
+import {getFacilityAddresses, getFacilityPhones, getOtherFacilityData} from 'selectors/facilityDataSelectors'
+
+const mapStateToProps = (state) => ({
```
ratneshraval: Nice clean containers.

config/routes.rb
Nasir2712: So, since we are using react router, the rails controller methods are just POST requests to get the desired response. If we were using rails routing then the above suggestion would have been used.

app/javascript/containers/facilityAddressContainer.js
Nasir2712: 👍

ratneshraval:
Closing in favor of #514 as same code was pulled in that PR.

PR #511, opened 2018-06-01
AARA-395: fix for blank page on RFA list

LeonidMarushevskyi:
- https://osi-cwds.atlassian.net/browse/AARA-395

- Rendering of RFA list was failing for objects which have empty address array in residence field. Example of residence object: 
{"addresses": [], "home_languages": [], "weapon_in_home": null, "directions_to_home": "", "body_of_water_exist": null, "residence_ownership": null, "physical_mailing_similar": null, "body_of_water_description": "", "others_using_residence_as_mailing": false, "other_people_using_residence_as_mailing": null}

- No

- No

- 88.24% Statements 
- 79.6% Branches 
-  85.79% Functions 
-  88.44% Lines


app/javascript/rfa_forms/rfa01a_list/applicationsListRow.js
```diff
@@ -13,8 +13,8 @@ const getPhoneNumber = (phonesList) => {
 }
 
 const getAddress = (address, key) => {
-  const residentialAddress = address.get(0)
-  return residentialAddress.get(key)
+  const residentialAddress = address && address.get(0)
+  return residentialAddress && residentialAddress.get(key)
```
anuroop15: good catch on address empty or not. But is residentialAddress check needed in return statement. As object will have street, city and zip in all cases.

test/javascript/rfa_forms/rfa01a_list/applicationsListRow_tests.js
```diff
@@ -65,3 +65,24 @@ describe('Verify Application List Rows', () => {
...
+    facilityId: undefined,
+    familyName: undefined,
+    applicationStatus: undefined,
+    applicationReceivedDate: undefined,
+    applicantsInfo: application,
+    applicationAddress: address
+  }
+  const AppListViewCard = shallow(<ApplicationsListRow {...props} />)
+  it('To Return table Row', () => {
+    expect(AppListViewCard.find('tr').length).toEqual(1)
```
anuroop15: good to have check on table with table header. 
ecpect(AppListViewCard.find('th').at(0)).toBe('FacilityID'). to make sure table header is rendered.

app/javascript/rfa_forms/rfa01a_list/applicationsListRow.js
LeonidMarushevskyi: It is needed. If address is empty array then residentialAddress will be undefined. Then on line 17 we will have "Cannot read property 'get' of undefined". You can reproduce this error using test which I wrote.

anuroop15: defaultProps should take of this scenario. Not sure why it's not working in this case.

test/javascript/rfa_forms/rfa01a_list/applicationsListRow_tests.js
LeonidMarushevskyi: Added test to check that default address is used when address supplied as empty array

PR #517, opened 2018-06-08
AARA-420: update PUT implementation

lauherkcambria:
- https://osi-cwds.atlassian.net/browse/AARA-420

-
implement update PUT call on rails side for front end consumption via a save button on tracking page

-no

-no

-n/a


lauherkcambria:
jenkins, retest this please.

lauherkcambria:
jenkins, retest this please.

app/controllers/rfa/tracking_controller.rb
```diff
@@ -13,14 +13,31 @@ def edit
...
+    render json: e.response, status: e.status
   end
 
-  def delete; end
-
   private
 
-  def tracking_params; end
+  def tracking_params
+    params.require(:tracking).permit(:id, :facility_name, :license_number, :rfa_1a_id,
```
anuroop15: @lauherkcambria: When I compare params permit with the tracking API, the structure was different. was it made intentionally??
for example, tracking_documents was mentions as array, but it's an object in API contract. Can I know why is that so??

app/javascript/rfa_forms/tracking/index.js
```diff
@@ -1,24 +1,62 @@
...
+          headerLabel='RFA Tracking'
+          pageHeaderButtons={
+            <Button
+              buttonId='saveProgress'
+              label='Save '
+              textAlignment='right'
+              onClick={this.saveProgress}
+            />}
+        />
+        <BreadCrumb
```
nasar-cambria: These changes regarding BreadCrumb is unrelated to the topic of the PR, it looks like an important change to merge in. Separating it into its own story would provide us flexibility in what we merge.

app/javascript/rfa_forms/tracking/index.js
```diff
@@ -1,24 +1,62 @@
...
+  }
 
   render () {
     return (
-      <CardsGroupLayout>
-        <div>tracking id: {this.state.tracking.id}</div>
-        <div>rfa_1a_id: {this.state.rfa_application.id}</div>
-      </CardsGroupLayout>
+      <div className='tracking_page'>
+        <PageHeader
```
nasar-cambria: Same thing here. Just calling it out so you can add this in the Acceptance criteria and it's clear to everyone that it's a part of this story.

app/javascript/rfa_forms/tracking/index.js
lauherkcambria: see other comment

app/javascript/rfa_forms/tracking/index.js
lauherkcambria: cool - i'm leaving this in as we discussed.it's a small addition that doesn't effect anything negatively,  and will ultimately need to be implemented. good thought, I'll add it to the acceptance criteria.

app/controllers/rfa/tracking_controller.rb
lauherkcambria: this is rails syntax for permitting parameters.

PR #521, opened 2018-06-11
Refactoring Children to match facility and complaints architecture

Nasir2712:
- n/a

- n/a

- no

- no

-n/a


app/controllers/facilities/children_controller.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class Facilities::ChildrenController < CalsBaseController
+  include Response
+
+  def index
+    @facility_children = child_helper.find_by_facility(params[:facility_id])
+    json_response @facility_children.nil? ? { "children": [] } : @facility_children
```
Nasir2712: will remove this condition handling, once api fixes it on their end.

PR #524, opened 2018-06-12
AARA-411: Spike: Editable grid/table implementation

anuroop15:
#### Which User story does this relate to (link)?
- https://osi-cwds.atlassian.net/browse/AARA-411

#### Brief feature description
- POC on having re-usable table for tracking table

#### Does this PR block/effect other stories?
- NA
 
#### Do any commands needs to be run in order to run the project with this PR?
-

#### Code coverage percent diff as a result of this PR?
-

PR #526, opened 2018-06-13
Refactor tracking documents

nasar-cambria:
- Creating a follow up story to @anuroop15 PR: https://github.com/ca-cwds/CALS/pull/524 to refactor editSaveToggle method.

Adding TrackingDocument component would help us pass props inside it and keep the toggle functionality on the index level rather than nesting it in TrackingTable page. 

Helpful Links: [Destructuring 1](http://www.deadcoderising.com/2017-03-28-es6-destructuring-an-elegant-way-of-extracting-data-from-arrays-and-objects-in-javascript/) [Destructuring 2](https://hacks.mozilla.org/2015/05/es6-in-depth-destructuring/)


nasar-cambria:
As discussed, I am closing this story for now as we mutually agreed to use the same approach as is in development.

PR #531, opened 2018-06-18
SEAR-92 Add Facility Name on facility profile

sunilb0575:
-https://osi-cwds.atlassian.net/browse/SEAR-92

-Facility Name is added on facility profile. 

- No

-No

-No


sunilb0575:
Yes, he also commented the same in JIRA.

PR #533, opened 2018-06-18
increase capybara wait time to 10 sec

anuroop15:
- NA
 
- Remove second applicant making an additional delete call making a timeout error on acceptance tests. So increasing the wait time to 10 seconds instead of 5 sec.

-

-

-


anuroop15:
tests are stable. Need to figure out why it needs 10 seconds to remove applicant, if this works.

PR #537, opened 2018-06-21
SEAR-249-CALS - Upgrade capybara_accessible to current version

Nasir2712:
- https://osi-cwds.atlassian.net/browse/SEAR-249

- add capybara accessible for accessibility testing for cals search.

- no

- bundle install

-N/A


spec/support/capybara.rb
```diff
@@ -48,15 +48,18 @@ def register_internet_explorer(capybara_config, selenium_browser)
 def registar_chrome(capybara_config, selenium_browser)
   capability_options = selenium_browser == :headless_chrome ? %w[headless disable-gpu no-sandbox] : []
 
-  capybara_config.register_driver selenium_browser do |app|
+  capybara_config.register_driver :accessible_selenium do |app|
```
ratneshraval: So before we were able to send `chrome` and `headless_chrome` to switch between the 2. Does that still work here?

spec/rails_helper.rb
```diff
@@ -29,6 +29,10 @@
     end
   end
 
+  config.around(:each, inaccessible: true) do |example|
```
ratneshraval: Let's create a shell story in RFA backlog to so that team can fix RFA tests and eventually get rid of this.

spec/support/capybara.rb
```diff
@@ -78,3 +81,18 @@ def registar_chrome(capybara_config, selenium_browser)
 end
 
 Capybara.app = Rack::Builder.parse_file(File.expand_path('../../../config.ru', __FILE__)).first
+
+module Capybara
+  module Accessible
+    class SeleniumDriverAdapter
+      def modal_dialog_present?(driver)
+        driver.browser.switch_to.alert
```
ratneshraval: What does this do?

spec/support/capybara.rb
Nasir2712: yes

spec/support/capybara.rb
Nasir2712: It is a setup to test the browser, the code actually sees if the browser that we pass in the ```modal_dialog_present``` check to see if a javascript alert can be created or not.

PR #541, opened 2018-06-22
fix loading spinner IE

Nasir2712:
-

-

-

-

-


ratneshraval:
@Nasir2712 this seems like a lot of code for us to manage specific IE 11 issue. Can you try with one of the library you mentioned?

Nasir2712:
closing this PR because another has been created.

PR #556, opened 2018-06-29
AARA-385:  Buttons extend beyond page header when display resolution is not at maximum

anuroop15:
- https://osi-cwds.atlassian.net/browse/AARA-385

-  RFA list view page header buttons to fit within pageheader

- NA

- NA
 
- NA


lauherkcambria:
@anuroop15 down to what screen size are we targeting? at 1024x768 the buttons start to blur- what does PO think? 
image

anuroop15:
@lauherkcambria I have adjusted the css to support 1024 resolution. I initially fixed it for 1200px (laptop) resolutions. Can you re-check the buttons now. I have fixed it.

PR #558, opened 2018-07-02
AARA_431 Applicant gender field should be marked as required

nasar-cambria:
- https://osi-cwds.atlassian.net/browse/AARA-431

- Applicant gender field should be marked as required

- No

- No

- NA


nasar-cambria:
@lauherkcambria - This error is in integration too, doesn't looks like it is related with this story. I can show it to @BinthuKennedyCWDS so we can have a new story logged in Jira.

PR #563, opened 2018-07-03
AARA-261: make RFA 01A side bar font same size as right side content font

lauherkcambria:
-https://osi-cwds.atlassian.net/browse/AARA-261

- change the side bar title font to be 2rem (same as title in cards) and font to be 1.5 rem (same as body font size) 

**NOTE**: this pr only addresses the font sizes, and not the upcoming design changes that may result in a revamp of the sidebar itself.

-no

-no

-n/a


lauherkcambria:
jenkins, retest this please

nasar-cambria:
screen shot 2018-07-09 at 8 59 21 am
@lauherkcambria - Font size is constant now which is good. One thing I noticed is that Applicant information is aligned right even on the desktop view and on re-size the text shifts down.

anuroop15:
@lauherkcambria  There is a small alignment issue occurred while fixing the font size. 
Also, links to be left aligned but it's not.

It is not fit under our our standard resolution of 1024*786px.

screen shot 2018-07-09 at 9 47 08 am

nasar-cambria:
re test this please

lauherkcambria:
@anuroop15 i've addressed comments

PR #572, opened 2018-07-09
SEAR-296 Add Params to Dora Endpoint

Nasir2712:
-https://osi-cwds.atlassian.net/browse/SEAR-296

-add query params

- no

-no

-no


ratneshraval:
This can be closed as only 1 shard is used for facilities index. If the number of shards are changed in future or the issue persist then we will have to revisit.

PR #576, opened 2018-07-10
Fix extra padding for allegations table in  IE

sunilb0575:
- Remove extra padding in allegations table of the complaints history section specific IE browser.

-No

-No

-No


sunilb0575:
There are still some issues with styles in IE11, will close this PR as this is stale and will try to fix the issue in new PR.

PR #578, opened 2018-07-10
remove session timeout redis

anuroop15:
- https://osi-cwds.atlassian.net/browse/AARA-246

- Session timeout depend on Perry instead of redis server

- NA
 
- NA

- NA


ratneshraval:
Please do not completely remove session timeout until there is another solution in place. Either leave it at 4 hours or restore it to previous value. Removing session timeout completely seems like a security problem.

anuroop15:
closing this branch as there is not resolution yet on browser inactivity timeout session expiration.

PR #583, opened 2018-07-16
SEAR-267 Application SHOULD NOT Require Exact Match of Facility Name in order to Show in Results

Nasir2712:
-https://osi-cwds.atlassian.net/browse/SEAR-267

- partial search results using the facility name and facility search

-n/a

-n/a

-


Nasir2712:
jenkins, retest this please

app/classes/elastic/query_builder.rb
```diff
@@ -3,15 +3,22 @@ class Elastic::QueryBuilder
...
     # each key is an attribute to query on
     query_hash.each do |k, v|
-      query_leaves << { v[:query_type].to_sym => { k.to_sym => v[:value] } }
+      query_leaves << name_and_other_queries(k, v)
     end
 
     return query_leaves
   end
 
+  def self.name_and_other_queries(k, v)
```
ratneshraval: I think better name is `query_by_type` or something like that.

app/javascript/search/common/commonUtils.js
```diff
@@ -27,6 +27,17 @@ export const checkForValue = (value) => {
   return value || undefined
 }
 
+export const checkAndSplitValue = (value) => {
+  if (value !== undefined || '') {
+    return value.split(' ').map((val) => {
```
ratneshraval: Let's use `filter` instead of `map + regex`

Nasir2712: done

app/classes/elastic/query_builder.rb
Nasir2712: done

app/javascript/search/common/commonUtils.js
```diff
@@ -27,6 +27,13 @@ export const checkForValue = (value) => {
   return value || undefined
 }
 
+export const checkAndSplitValue = (value) => {
+  if (value !== undefined || '') {
+    const filteredSet = ['*', '']
```
ratneshraval: 👍

PR #585, opened 2018-07-17
Search Dictionaries Refactor

Nasir2712:
- n/a

- to fetch dictionaries GET request should be used instead of a POST request because no params are being posted.

- no

- no

- no


app/javascript/sagas/fetchSearchDictionariesSaga.js
```diff
@@ -9,7 +9,7 @@ import {SEARCH_DICTIONARIES_FETCH} from 'constants/actionTypes'
 export function * fetchSearchDictionaries (action) {
   try {
     const url = '/search/user_and_dictionaries'
-    const response = yield call(fetchRequest, url, 'POST')
+    const response = yield call(fetchRequest, url, 'GET')
```
ratneshraval: So while you're here, there should be 2 different calls. One for user and one for search related dictionaries.

Nasir2712: done!

config/routes.rb
```diff
@@ -9,17 +9,18 @@
   end
 
   resources :search do
-    collection { post :user_and_dictionaries}
+    collection do
+      get :search_dictionaries
```
ratneshraval: This is fine for now but there should be separate route for getting dictionaries when we need to use it outside of search, in other areas.

PR #596, opened 2018-07-25
set up redux dev tools for easier debugging

Nasir2712:
- n/a

- set up redux dev tools.

-n/a

-n/a

-n/a


ratneshraval:
Looks like it is only setup in dev mode/env, right? Disabled in production?

PR #598, opened 2018-07-26
SEAR-292-Children Sort order

Nasir2712:
-https://osi-cwds.atlassian.net/browse/SEAR-292

- sorting ability for children via date of placement

- n/a

-n/a

-n/a


app/javascript/search/common/commonUtils.js
```diff
@@ -71,6 +71,12 @@ export const checkforDateOrNa = (date) => {
   }
 }
 
+export const sortbyDate = (a, b) => {
```
ratneshraval: So why do we need to write our own function for date sorting?

Nasir2712: the default sorting was not helping

ratneshraval: My guess is the data type is string hence default sorting won't work.
You can tell react-table about the data type or convert it at cell level and hopefully the default sort will work after that.

Nasir2712: It works for all strings but doesn't work for the date. Like for example, default sorting works for first name and last name, but not for the date. Tried converting to a date object still does not work.

Nasir2712:
jenkins, retest this please

PR #599, opened 2018-07-27
Update Cancel Functionality to handle setState asynchronous calls

anuroop15:
- https://osi-cwds.atlassian.net/browse/AARA-505

- Updated cancelProgress method to handle setState method

- AARA-505.

- NA

-


lauherkcambria:
nice! :D can we address the handle clear on conditional change issue here as well? 
image

anuroop15:
@lauherkcambria will address handle clear on separate PR.

PR #600, opened 2018-07-27
AARA-506: Handle conditional methods to set default values

anuroop15:
- https://osi-cwds.atlassian.net/browse/AARA-506

- Using delete operation to remove objects on yes no selection
 
- NA

-  NA

- NA


app/javascript/rfa_forms/rfa02_edit_view/outOfStateRegistryCheck.jsx
```diff
@@ -20,6 +20,9 @@ export default class OutOfStateRegistryCheck extends React.Component {
   }
   handleOutOfStateRegistry (key, value, index) {
     let outOfStateRegistryList = this.props.outOfStateRegistryList
+    if (key === 'is_registry_maintained_by_state' && value === 'false') {
+      outOfStateRegistryList = outOfStateRegistryList.deleteIn(['state_registries', index, 'registry_info'])
```
lauherkcambria: same as above

app/javascript/rfa_forms/rfa02_edit_view/exemptions.jsx
```diff
@@ -14,6 +14,12 @@ export default class Exemptions extends React.Component {
   exemptionHandler (key, value, subKey) {
     let exemptionList = this.props.exemptionList
     let newExemptionList
+    if (key === 'is_requested' && value === 'false') {
+      newExemptionList = exemptionList.set(key, value)
+      newExemptionList = newExemptionList.delete('date')
+      newExemptionList = newExemptionList.delete('notes')
+      newExemptionList = newExemptionList.delete('approval')
```
lauherkcambria: can we set default props here instead of immutable delete ?

PR #603, opened 2018-07-30
SEAR-345-MFA-Error when logout from Facility Search Page

Nasir2712:
-https://osi-cwds.atlassian.net/browse/SEAR-345

- should be able to logout and not throw any errors

- n/a

-n/a
-n/a


Nasir2712:
retest this please

PR #609, opened 2018-08-06
AARA-497: Submit functionality fixed to not create duplicates of applicants, minor children and other adults

LeonidMarushevskyi:
- https://osi-cwds.atlassian.net/browse/AARA-497

- Previously submit button functionality was implemented on Rails side and was invoking saveProgress before submit. This approach is not correct because when submit fails with some error, application was saved to database but frontend state has no knowledge of current application state. In this PR Submit functionality moved to Javascript to solve described issue.

- No

- No

- No diff


LeonidMarushevskyi:
jenkins retest this please

LeonidMarushevskyi:
Changes from this PR now completely included into PR https://github.com/ca-cwds/CALS/pull/610. So closing this one.

PR #610, opened 2018-08-07
AARA-512, AARA-497: submit button 400 fix, plus other bug fixes that didn't allow me to test this bug

lauherkcambria:
- https://osi-cwds.atlassian.net/browse/AARA-497 
- https://osi-cwds.atlassian.net/browse/AARA-512

this commit addresses the following issues: 
 - properly configure submit button on b and c to submit a application
 - fix save progress on rails side to be on react side
 - fix submit enabled bug that on refresh allowed an inproperly saved application submitabble.
 - fix specs up test and spec cov

**Note:** These bugs were discovered upon trying to test the fix for this reported bug, and were needed to be addressed in order to make progress and fix/test this bug. 

- no

- no 

- 90.04% Statements 
- 81.51% Branches 
- 87.36% Functions 
- 90.25% Lines 
-spec : 90.5%


lauherkcambria:
jenkins, retest this please

lauherkcambria:
jenkins, retest this please.

LeonidMarushevskyi:
An offline comment regarding asynchronous invocations of HHTP calls to saveProgress and submit was addressed. Now on submit those calls will wait for the response of the first call before start next.

Submit buttons now works as expected in all RFA 1 forms (a, b, c). The only thing missing is conditional logic between forms which will prevent to submit an application if other forms don't have all required fields. This functionality is out of scope for current PR.

app/javascript/rfa_forms/rfa01b_edit_view/index.js
```diff
@@ -145,9 +137,15 @@ export default class Rfa01bEditView extends React.Component {
...
   setApplicationState (key, value) {
-    let newState = Immutable.fromJS(this.state)
-    newState = newState.setIn(['application', key], value)
-    this.setState(newState.toJS())
+    let newStateApplication = Immutable.fromJS(this.state.application).set(key, value)
+    this.setState({
+      application: newStateApplication.toJS(),
+      disableSubmit: !this.validateAllRequiredForSubmit(newStateApplication.toJS())
+    })
+    this.setState({
```
anuroop15: why do we need two setState method if we can handle disableSubmit before setting new state?

app/javascript/helpers/validator.js
```diff
@@ -17,7 +17,9 @@ export default class Validator {
       isNotInTheFuture: this.isNotInTheFuture,
       isBeforeOtherDate: this.isBeforeOtherDate,
       is10digits: this.is10digits,
-      isValidDate: this.isValidDate
+      isValidDate: this.isValidDate,
+      isRequiredArray: this.isRequiredArray,
```
anuroop15: Are these validations 'isRequiredArray & isRequiredIfArray" needed?? Any specific need these validation fulfill in code?

app/javascript/rfa_forms/rfa01b_edit_view/index.js
lauherkcambria: do you have latest changes?

app/javascript/helpers/validator.js
lauherkcambria: I left them in the validator because we will need to use these methods, as I'm sure we'll have to validate an array with a conditional at some point. but you're right in that we don't need to add them in the file

lauherkcambria: updated

nasar-cambria:
screen shot 2018-08-07 at 5 10 08 pm

lauherkcambria:
jenkins, retest this please

lauherkcambria:
jenkins, retest this please

lauherkcambria:
@nasar-cambria  thanks for finding this issue, i've created a PR for it : https://github.com/ca-cwds/CALS/pull/614

PR #618, opened 2018-08-10
SEAR-228 Update Facility Profile Labels

Nasir2712:
-https://osi-cwds.atlassian.net/browse/SEAR-228

- Update facility profile labels


app/javascript/facility/facilityChildren.jsx
```diff
@@ -28,10 +28,10 @@ export default class FacilityChildren extends React.Component {
               defaultPageSize={children.length}
               showPagination={false}
               sortable={true}
-              defaultSorted={[{
+              defaultSorted={children.length > 0 ? [{
```
ratneshraval: No test for this condition?

Nasir2712: So testing sorting is kinda tricky using react table,  the way we do it in RFA, is give a selector with index for a given html element, tried the same approach for testing here, but that didn't work because the selector's index is not sorted.

PR #653, opened 2018-11-08
CANS-540, FIT team Added health check

AVandana-OSI:
https://osi-cwds.atlassian.net/browse/CANS-540

Added health check using the System Information GEM, which follows the standards as follows
https://github.com/ca-cwds/development-practices/blob/master/health_checks.md
which includes redis, Perry, CALS API, Dora API, GEO API were CALS dependent on.

No

No

-


AVandana-OSI:
retest this please

SashaKyz:
retest this please

SashaKyz:
retest this please

README.md
```diff
@@ -22,7 +22,7 @@ Make sure XCode is installed.
 ## Initialization
 
 run the command:
-`yarn run dev-build`
+`yarn dev`
```
edgibbs: So the command for `dev-build` still exists and the `dev` command runs foreman instead, so I'd need to have more context on whether it's the better default.

ratneshraval: `yarn dev` is better as it runs both rails and webpack dev server in hot mode.

AVandana-OSI: @edgibbs Yarn run dev-build was not starting the project locally, Yarn dev runs perfectly. Had a conversation with Nesh about the changing the read me file from `yarn run dev-build` to `yarn run` he is ok with it.

PR #656, opened 2018-11-28
CODE-80 enabling jsx-a11y eslint plugin and fixes

NagaPradeepGajula-osi:
- https://osi-cwds.atlassian.net/browse/CODE-80

- This PR involves enabling jsx-a11y eslint plugin and fix interactive-supports-focus, no-static-element-interactions, click-events-have-key-events issues

-

-

-


app/javascript/components/rfa_forms/countyUseOnlyCard.js
```diff
@@ -14,7 +14,8 @@ const CountyUseOnlyCard = ({
       <div className='county_use_only_card'>
         <div id='CountyUseOnlySection'
           onClick={() => setFocusState('CountyUseOnlySection')}
-          className={getFocusClassName('CountyUseOnlySection') + ' ' + 'card phone-section double-gap-top active-bar'}>
+          className={`${getFocusClassName('CountyUseOnlySection')} ` + `card phone-section double-gap-top active-bar`}
```
ratneshraval: You don't need plus if you're doing string interpolation
```
className={`${getFocusClassName('CountyUseOnlySection')} card phone-section double-gap-top active-bar`}
```

app/javascript/rfa_forms/rfa01a_edit_view/FosterCareHistoryCard.jsx
```diff
@@ -8,7 +8,8 @@ export default class FosterCareHistoryCard extends React.Component {
     return (
       <div className='foster_care_history_cards'>
         <div id='FosterCareHistorySection' onClick={() => this.props.setFocusState('FosterCareHistoryCard')}
-          className={this.props.getFocusClassName('FosterCareHistoryCard') + ' ' + 'card phone-section double-gap-top active-bar'}>
+          className={`${this.props.getFocusClassName('FosterCareHistoryCard')} ` + `card phone-section double-gap-top active-bar`}
```
ratneshraval: Same string interpolation comment as above

app/javascript/rfa_forms/rfa01a_edit_view/applicantMaritalHistoryCardGroup.js
```diff
@@ -111,16 +111,17 @@ export default class ApplicantMaritalHistoryCardGroup extends React.Component {
...
           <div className='card-header'>
             <span>Applicants Marital / Domestic Partnership History</span>
           </div>
           <div className='card-body'>
             {
               applicantMaritalHistories.map((applicantMaritalHistory, index) => {
                 return (
-                  <div key={'formerSpouse' + '[' + index + ']'} className='row list-item' >
-                    <div> <a onClick={(event) => this.onMaritalHistoryClickClose(index)} className='pull-right remove-btn'>Remove</a>
+                  <div key={`${'formerSpouse' + '['}${index}]`} className='row list-item' >
```
ratneshraval: Will this work? 
```
`formerSpouse[${index}]`
```

app/javascript/components/rfa_forms/countyUseOnlyCard.js
NagaPradeepGajula-osi: Ah! Good catch they were auto fixes by eslint. Thanks for pointing it out, I will make the change.

app/javascript/rfa_forms/rfa01a_edit_view/FosterCareHistoryCard.jsx
NagaPradeepGajula-osi: same here.

app/javascript/rfa_forms/rfa01a_edit_view/applicantMaritalHistoryCardGroup.js
NagaPradeepGajula-osi: it should work let me give it a try.

app/javascript/components/rfa_forms/countyUseOnlyCard.js
Nasir2712: I think if you are going to make that change then you have to make that change everywhere. This change is required in a lot of files. Do you know an automatic fix for this instead of fixing it manually?

NagaPradeepGajula-osi: Just doing it here as I touched it for jsx-a11y purpose. I am not going to fix any others. And there isn't an autoFix for it that I know but `no-useless-concat` kind of does some of it.

Nasir2712: Did you make this file change manually?

sunilb0575: There are few more places where string interpolation needs to be corrected.
For example: 
[https://github.com/ca-cwds/CALS/blob/CODE-80-enable-jsx-a11y-eslint-plugin/app/javascript/rfa_forms/rfa01a_edit_view/applicantCard.jsx#L29](url)

When and how are we planning to do that?

NagaPradeepGajula-osi: These are not reported by either Eslint or Code Climate. Is this effecting any of the existing code?
And I believe this is upto your team to fix them if needed.

sunilb0575: As you adjusted those incorrect string interpolation in two place, I thought we might also plan to fix rest of them. 
Ok can be in done in separate PR.

NagaPradeepGajula-osi: I  am ok to help with it later. But for now that is not the scope of my story.

PR #662, opened 2018-12-06
Set user_details in session if nil

ratneshraval:
- SEAR-466

- Shared redis session. Setting session user details if not present


ratneshraval:
retest this please

PR #684, opened 2019-01-25
Override magic numbers rule in tests as they are absolutely fine in tests

NagaPradeepGajula-osi:
- No story

- Before fix (90) warnings. After fix (16) warnings left from .js files

-


NagaPradeepGajula-osi:
retest this please

PR #716, opened 2019-04-27
Fit 399 integrate firefox testing

money:
- https://osi-cwds.atlassian.net/browse/FIT-399

- Run the current acceptance tests in firefox for CALS App

- `spec-end2end-acceptance-headless-firefox`
- `spec-end2end-acceptance-firefox`
**make sure to have the environment variables and VPN access if running against preint or integration**

## OH BOY
Lots of cleanup work is needed in this repo relating to Rubocop syntax standards.


money:
Jenkins retest this please

.rubocop.yml
```diff
@@ -2,10 +2,43 @@ AllCops:
   TargetRubyVersion: 2.4.2
```
ratneshraval: Should this be changed to 2.5.1 as well?

.rubocop.yml
```diff
@@ -2,10 +2,43 @@ AllCops:
...
 #   Excludes:
     # - app/javascript
+Lint/AmbiguousBlockAssociation:
```
ratneshraval: Are all these `false` rules being violated in current cals ?

.rubocop.yml
money: Should be, but I would rather do it outside story incase there are breaking changes that need to be addressed.

.rubocop.yml
money: I don't think so, I really didn't dig into it.

ratneshraval: In that case are we turning them off for all repos?

money: Yes. Can create stories to do the cleanup work.

NagaPradeepGajula-osi: > In that case are we turning them off for all repos?

No, we will be not turning these off on other repos.

ratneshraval: Then why are we turning them off here? If there isn't a violation, the rule should not be turned off.
If there is a violation and rule should be checked, then maybe add a story for each rule so someone can pick it up later.

money: @ratneshraval 
I am misunderstanding your initial question.

_are all these `false` rules being violated?_

If you run rubocop on the master branch, you will see that they are not *false* rules that are being violated.  Because the violations are valid and need to be addressed, I disabled them in order to move the story forward.

The total count was around 800 and they vary in complexity. For example, refactoring methods with too many lines.

The fixes in here were the low hanging fruits that rubocop fixes for you.

So to your latter point: 
_If there is a violation and rule should be checked, then maybe add a story for each rule so someone can pick it up later._ 
That is exactly what the goal is.

I hope this clarifies.  If you need further clarification would be happy to provide.

edgibbs: So started refactoring and found out: "Here be dragons"

```
                                                      ____________
                                 (`-..________....---''  ____..._.-`
                                  \\`._______.._,.---'''     ,'
                                  ; )`.      __..-'`-.      /
                                 / /     _.-' _,.;;._ `-._,'
                                / /   ,-' _.-'  //   ``--._``._
                              ,','_.-' ,-' _.- (( =-    -. `-._`-._____
                            ,;.''__..-'   _..--.\\.--'````--.._``-.`-._`.
             _          |\,' .-''        ```-'`---'`-...__,._  ``-.`-.`-.`.
  _     _.-,'(__)\__)\-'' `     ___  .          `     \      `--._
,',)---' /|)          `     `      ``-.   `     /     /     `     `-.
\_____--.  '`  `               __..-.  \     . (   < _...-----..._   `.
 \_,--..__. \\ .-`.\----'';``,..-.__ \  \      ,`_. `.,-'`--'`---''`.  )
           `.\`.\  `_.-..' ,'   _,-..'  /..,-''(, ,' ; ( _______`___..'__
                   ((,(,__(    ((,(,__,'  ``'-- `'`.(\  `.,..______   SSt
                                                      ``--------..._``--.__

```

PR #732, opened 2019-07-11
SEAR-631 Display Complaint Last Update Date in Facility Profile when Complaints are found

AlexanderSerbin:
https://osi-cwds.atlassian.net/browse/SEAR-631

fixed bug when batch is failured and allegations are showing up


app/javascript/selectors/facilityDataSelectors.js
```diff
@@ -175,8 +175,7 @@ export const getLastBatchSuccess = (state) => {
 export const getLastBatchTimestamp = (state) => {
   if (state.facilityComplaints && lastBatchResultSelector(state) &&
     lastBatchResultSelector(state).last_batch_timestamp) {
-    return checkforDateOrNa(lastBatchResultSelector(state).last_batch_timestamp) +
-    lastBatchResultSelector(state).last_batch_timestamp.substr(positionOfSpaceInTheTimestamp)
+    return lastBatchResultSelector(state).last_batch_timestamp
```
edgibbs: So I'm assuming the format coming back from the API has changed so it doesn't need to format it as much?

AlexanderSerbin: Yeah. That's correct. There is a new version of API returning timestamp in proper format

